PR #127, opened 2021-01-26
Fix unstable diff in ControlTemplates.json

lesaltzm:


src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -34,7 +34,8 @@ internal static partial class SourceSerializer
         // 12 - Moved Resources.json, move volatile rootpaths to entropy
         // 13 - Control UniqueIds to Entropy
         // 14 - Yaml DoubleQuote escape
-        public static Version CurrentSourceVersion = new Version(0, 14);
+        // 15 - Use dictionary for templates
+        public static Version CurrentSourceVersion = new Version(0, 15);
```
MikeStall, line 38: nit: is this actually a format update? Ie, do versions saved in 14 no longer load?

MikeStall (approved):

src/PAModel/Serializers/SourceSerializer.cs
lesaltzm: It is, yeah. I missed it originally, but the List<KVP> gets written as [{Key: Foo, Value: {....}}, {Key: Bar, Value: {....}}....] whereas the dict gets written as { Foo: {....}, Bar: {...} }.

Issue #128, opened 2021-01-27
Test failures on Linux and MacOS


lesaltzm:

MikeStall:
Issue #77 may also be related - newline encoding.

SeaDude:
Test failure output. 
Ubuntu 20.10
```
user5@system5:~/Documents/powerapps/packUnpack/PowerApps-Language-Tooling$ ./build ci
ci: Starting... (ci)
ci: squeaky-clean: Starting...
Cleaning directory: /home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/bin
Cleaning directory: /home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/obj
Cleaning directory: /home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/pkg
ci: squeaky-clean: Succeeded. (1.07 ms)
ci: clean: Starting...
targets: dotnet clean /home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/src/PASopa.sln --configuration Debug /clp:verbosity=minimal /flp:Verbosity=normal;LogFile=/home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/obj/logs/clean-Debug.log /flp3:PerformanceSummary;Verbosity=diag;LogFile=/home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/obj/logs/clean-Debug.diagnostics.log -p:GitExists=true /nologo
ci: clean: Succeeded. (1.49 s)
ci: restore: Starting...
targets: dotnet restore /home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/src/PASopa.sln /clp:verbosity=minimal /flp:Verbosity=normal;LogFile=/home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/obj/logs/restore-Debug.log /flp3:PerformanceSummary;Verbosity=diag;LogFile=/home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/obj/logs/restore-Debug.diagnostics.log -p:GitExists=true /nologo
  Determining projects to restore...
  Restored /home/user5/Documents/powerapps/packUnpack/PowerApps-Language-Tooling/src/PASopa/PASopa.csproj (in 1.23 sec).
[...]

PR #132, opened 2021-01-27
Enabling cross-platform builds and tests.

MikeStall:
Still have some known test failures to address:
https://github.com/microsoft/PowerApps-Language-Tooling/issues/128


src/PAModelTests/UtilityTests.cs
```diff
@@ -22,7 +23,7 @@ public void TestEscaping(string unescaped, string escaped)
         public void TestNotEscaped()
         {
             // Not escaped.
-            var a = "0123456789AZaz[]_. \\";
```
MikeStall, line 25: What's the impact of removing / ?

targets/Program.cs
```diff
@@ -131,7 +131,7 @@ static void CleanDirectory(string directoryPath)
                 }
             }
             catch (AccessViolationException) { /* swallow */ }
-            // catch (Exception ex) { Console.WriteLine(ex.Message); }
+            catch (IOException) { Directory.Delete(directoryPath, true); /* try again */ }
```
MikeStall, line 134: What's this?

src/PAModel/Utility/Utility.cs
```diff
@@ -211,7 +211,7 @@ private static bool DontEscapeChar(char ch)
                 (ch == '_') ||
                 (ch == '.') ||
                 (ch == ' ') || // allow spaces, very common.  
-                (ch == '\\'); // Allow directory separators. 
+                (ch == Path.DirectorySeparatorChar); // Allow directory separators.
```
MikeStall, line 214: This doesn't feel right... escaping should be agnostic to OS. 
IE, what if we unpack on windows, check into github, and then pack on Linux?

src/PAModel/Checksum/ChecksumMaker.cs
```diff
@@ -290,7 +291,7 @@ private static void ChecksumJson(Context ctx, IncrementalHash hash, JsonElement
                     else if (ctx.IsXmlDoubleEncoded && !string.IsNullOrWhiteSpace(str))
                     {
                         var parsedXML = XDocument.Parse(str);
-                        var xmlString = parsedXML.ToString(SaveOptions.None);
+                        var xmlString = parsedXML.ToString(SaveOptions.None).Replace("\r\n", "\n");
```
MikeStall, line 294: this looks like the critical spot for checksum? catching the double escaped xml?

lesaltzm: yup

targets/Program.cs
lesaltzm: whoops, something I was testing out, I was running into an issue where some file handles were sticking around longer than I wanted. not actually needed, I think

src/PAModel/Utility/Utility.cs
lesaltzm: Right, yeah. This works for linux->linux round trip, and windows->windows. I think the solution for linux->windows is we need to escape both directory separators, probably.

PR #137, opened 2021-02-01
Fix roundtrip failures for Legacy DataTable Control

### LINKED ISSUE #120: Warning PA2001: Checksum mismatch. File Controls\194.json checksum does not match on extract
What be the cause of the warning:
```
Warning PA2001: Checksum mismatch. File Controls\194.json checksum does not match on extract
```
I also get messages like the following:
```
FAIL: hash mismatch: Controls\145.json
FAIL: hash mismatch: Controls\194.json
FAIL: hash mismatch: Controls\22.json
FAIL: hash mismatch: Controls\3.json
FAIL: hash mismatch: Controls\88.json
FAIL: hash mismatch: Properties.json
FAIL: hash mismatch: References\Templates.json
FAIL!!
```

This happens on first extract where there were no existing files extracted.

lesaltzm:
The CustomControlJson can vary slightly between columns for the legacy data table control. However, it's safe to just pick any individual json and use it for all the columns.

fixes #120

### DIFF
--- a/src/PAModel/IR/IRStateHelpers.cs
+++ b/src/PAModel/IR/IRStateHelpers.cs
@@ -401,6 +401,11 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             return property;
         }
 
+
+        // These two functions (split and recombine CustomTemplates) are responsible for handling the legacy DataTable control's
+        // CustomControlDefinitionJson. This JSON contains a stamp of which version it was created in, but that is the only difference
+        // As such, they were safe to move to Entropy. If entropy is removed, the one in the TemplateStore works fine for all instances
+        // of the control.
         private static void SplitCustomTemplates(Entropy entropy, ControlInfoJson.Template controlTemplate, string controlName)
         {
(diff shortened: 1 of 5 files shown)


src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -396,5 +400,29 @@ private static ControlInfoJson.RuleEntry GetPropertyEntry(ControlState state, Er
...
+        {
+            if (controlTemplate.Id != "Microsoft.PowerApps.CustomControlTemplate")
+                return;
+
+            entropy.AddDataTableControlJson(controlName, controlTemplate.CustomControlDefinitionJson);
+        }
+
+        private static void RecombineCustomTemplates(Entropy entropy, ControlInfoJson.Template controlTemplate, string controlName)
+        {
+            if (controlTemplate.Id != "Microsoft.PowerApps.CustomControlTemplate")
```
MikeStall, line 419: nit: Can this share a string constant with line 411?

MikeStall (approved):

Issue #94, opened 2021-01-06
oom on msapps with large files


MikeStall:
The basic problem is that we load the entire msapp into memory (CanvasDocument). 

And then oom at places like this:

```
                        case FileKind.AppCheckerResult:
                            var appChecker = Encoding.UTF8.GetString(entry.ToBytes());
                            app._entropy.AppCheckerResult = appChecker;
                            break;
```

We should calculating this in a streaming way rather than ToBytes(). 
(sarif file can be huge)

We should scrutinize usage of FileEntry.ToBytes, and probably need to be lazier and have documents backed by disk.

landonarogers:
@MikeStall Is this issue still in progress?

PR #140, opened 2021-02-02
Move Sarif file and other deletables to Entropy/, and Themes to src/

### LINKED ISSUE #121: Make AppCheckerResult.sarif available
#72 moved AppCheckerResult.sarif to Entropy file.  We want to use this Azure DevOps extension:
https://marketplace.visualstudio.com/items?itemName=sariftools.scans
...to be able to review app checker results during the Pull Request process.  The extension makes the contents of *.sarif files readable on the individual build page in Azure DevOps by adding a Scans tab.  Want to make it part of our process to review the Scans tab as part of a PR.  We require builds when PRs are submitted or when the source branch for the PR is updated.  To use the extension for this review process, the extension requires one can add grab a *.sarif file and publish it to an Azure DevOps artifact.

lesaltzm:
fixes #121

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -51,6 +51,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         internal TemplatesJson _templates;
         internal ThemesJson _themes;
         internal ResourcesJson _resourcesJson;
+        internal AppCheckerResultJson _appCheckerResultJson;
 
         // Environment-specific information
         // Extracted from _properties.LocalConnectionReferences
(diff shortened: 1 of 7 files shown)


MikeStall (approved):

src/PAModel/CanvasDocument.cs
```diff
@@ -51,6 +51,7 @@ public class CanvasDocument
         internal TemplatesJson _templates;
         internal ThemesJson _themes;
         internal ResourcesJson _resourcesJson;
+        internal AppCheckerResultJson _appCheckerResultJson;
```
MikeStall, line 54: Per #94, should we just keep .sarif it it's own \other\appcheckecer.sarif file? May get too large to put into entropy. 
And then the guidance is that  all of \other can get overwritten in a merge.

lesaltzm: Discussed offline, this is pretty much our best option for now.

MikeStall (approved):

PR #151, opened 2021-02-10
Stamp git hash and local build in Checksum.json

### LINKED ISSUE #149: build pipleine should stamp git hash
For diagnostics, particularly when troubleshooting a specific source format error, it can be very useful to know which version exactly of the alm tool produced it. 

The build pipeline should stamp the git commit hash, so that the ALM tool can then stamp it into the entropy\checksum.json.

lesaltzm:
fixes #149 

image

### DIFF
--- a/targets/Program.cs
+++ b/targets/Program.cs
@@ -29,7 +29,7 @@ namespace targets
             try 
             {
                 RootDir = Read("git", "rev-parse --show-toplevel", noEcho: true).Trim(); 
-                gitHash = Read("git", "rev-parse --short HEAD", noEcho: true).Trim();               
+                gitHash = Read("git", "rev-parse HEAD", noEcho: true).Trim();               
             }
             catch
             {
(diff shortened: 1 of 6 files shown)


targets/Program.cs
```diff
@@ -23,11 +24,12 @@ class Program
         static void Main(string[] args)
         {
             
-            string RootDir = "";
+            string RootDir = "", gitHash = "";
             bool gitExists = true;
             try 
             {
                 RootDir = Read("git", "rev-parse --show-toplevel", noEcho: true).Trim(); 
+                gitHash = Read("git", "rev-parse --short HEAD", noEcho: true).Trim();
```
lesaltzm, line 32: do we want the short or long hash?

MikeStall: May as well be long hash since this is all tooling.

MikeStall (approved):

PR #152, opened 2021-02-10
Fix relative path logic

### LINKED ISSUE #150: Error   PA3002: Format is not supported. Can't find CanvasManifest.json file - is sources an old version?
A week ago 

- Cloned the repo ,` build ci` 
- Downloaded and unpacked our 'App (Production).msapp'
- I successfully edited the source and remove an erronous sql  connection
- packed the app and opened it in our production environment, to check an annoying error "need to share the SQL Server connection " has gone.
- It had , so happy days!

Today I thought I would repeat the process  and document what I did but hit the issue below. Now this  is weird as nothing has changed incl the file I downloaded last week ( before any edits) , I am even using the same commands as found in my  Ps7 windows terminal, command history.

PASopa issue
 
Things i have tried ... 
I have pulled the latest source and rebuilt so now using 0.18 of the converter.

lesaltzm:
fixes #150

### DIFF
--- a/src/PAModel/Utility/Utility.cs
+++ b/src/PAModel/Utility/Utility.cs
@@ -144,10 +144,40 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         // basePath:     C:\foo
         // full: c:\foo\bar\hi.txt
         // returns "bar\hi.txt"
+
+        // Net Core 2.1 provides a Path.GetRelativePath api
+        // but since we target netstandard, we can convert to URIs and make the relative path from that
+        // see https://stackoverflow.com/questions/275689/how-to-get-relative-path-from-absolute-path
         public static string GetRelativePath(string full, string basePath)
         {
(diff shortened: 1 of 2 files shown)


src/PAModel/Utility/Utility.cs
```diff
@@ -144,10 +144,40 @@ public static JsonElement ToJson(this ZipArchiveEntry e)
         // basePath:     C:\foo
         // full: c:\foo\bar\hi.txt
         // returns "bar\hi.txt"
+
+        // Net Core 2.1 provides a Path.GetRelativePath api
+        // but since we target netstandard, we can convert to URIs and make the relative path from that
+        // see https://stackoverflow.com/questions/275689/how-to-get-relative-path-from-absolute-path
         public static string GetRelativePath(string full, string basePath)
```
MikeStall, line 151: Can you add a test case?

MikeStall (approved):

PR #165, opened 2021-02-25
better hash diagnostics.

MikeStall:
No change to actual hash algorithm.

### DIFF
--- a/src/PAModel/Checksum/ChecksumMaker.cs
+++ b/src/PAModel/Checksum/ChecksumMaker.cs
@@ -49,6 +49,15 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             return checksumMaker.GetChecksum();
         }
 
+        public void AddFile(string filename, byte[] bytes)
+        {
+            var key = ChecksumFile<Sha256HashMaker>(filename, bytes);
+            if (key != null)
+            {
+                _files.Add(filename, key);
+            }
+        }
+
         // Return null if not checksumed
         // T is the checksum aglorithm to use  - this allows passing in a debug algorithm. 
         internal static byte[] ChecksumFile<T>(string filename, byte[] bytes)
@@ -70,15 +79,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             {
                 // $$$ hash this?
                 return bytes;
-            }
-        }
-
-        public void AddFile(string filename, byte[] bytes)
-        {
-            var key = ChecksumFile<Sha256HashMaker>(filename, bytes);
-            if (key != null)
-            {
-                _files.Add(filename, key);
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/PAModel/Checksum/ChecksumMaker.cs
```diff
@@ -157,12 +163,12 @@ private void AddJsonFile(string filename, byte[] bytes)
 
                 var ctx = new Context { Filename = filename };
 
-                using (var hash = IncrementalHash.CreateHash(HashAlgorithmName.SHA256))
+                using (var hash = new T())
```
MikeStall, line 166: We can pass in a the real hash alg (sha256) or a "debug" one (which just collects the normalized values)

src/PAModel/Checksum/IHashMaker.cs
```diff
@@ -0,0 +1,177 @@
...
+            _hash.AppendData(value);
+        }
+
+        public void AppendPropName(string name)
+        {
+            this.AppendData(name);
+        }
+
+        public void AppendStartObj()
+        {
```
MikeStall, line 66: Worth noting this are all empty in the real checksum. We may want to consider hashing something here.

src/PAModel/MsAppTest.cs
```diff
@@ -86,85 +86,66 @@ public static bool Compare(string pathToZip1, string pathToZip2, TextWriter log)
...
             return false;
         }
 
+        // Compare the debug checksums. 
         // Get a hash for the MsApp file.
         // First pass adds file/hash to comp.
         // Second pass checks hash equality and removes files from comp.
         // AFter second pass, comp should be 0. any files in comp were missing from 2nd pass.
-        public static string Test(string pathToZip, TextWriter log, Dictionary<string,string> comp, bool first)
+        public static void DebugChecksum(string pathToZip, TextWriter log, Dictionary<string,byte[]> comp, bool first)
```
MikeStall, line 106: The real value of the change is making this debug function behave *exactly* the same as the checksum function. It shares the same json normalizing, double-escaping logic, etc. 
This runs the same checksum algorithm, but with a json writer (that captures the full output) rather than a sha256 incremental hash.

src/PAModel/Checksum/ChecksumMaker.cs
```diff
@@ -47,33 +49,36 @@ public static (string wholeChecksum, Dictionary<string, string> perFileChecksum)
...
             {
-                AddJsonFile(filename, bytes);
+                var key = ChecksumJsonFile<T>(filename, bytes);
+                return key;
             }
             else
             {
-                _files.Add(filename, bytes);
+                // $$$ hash this?
+                return bytes;
```
MikeStall, line 72: This may be a separate bug - we don't appear to hash binary files.

lesaltzm (approved):

PR #167, opened 2021-02-26
Make deterministic ordering

### LINKED ISSUE #166: ensure that yaml properties are strongly ordered [noisy diffs]
The yaml properties are bucketed into groups, but the specific order within a bucket is not canonical. 

this is technically not breaking, but will lead to noisy diffs.

MikeStall:
Fix #166

Since we can read any order, this is not a breaking change, so we don't need to update the source format.

### DIFF
--- a/src/PAModel/PAConvert/PAWriterVisitor.cs
+++ b/src/PAModel/PAConvert/PAWriterVisitor.cs
@@ -47,13 +47,14 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             node.Name.Accept(this, context);
             context._yaml.WriteStartObject(context._sb.ToString());
 
-            
-            foreach (var func in node.Functions)
+            // Ensure the order we write properties is canonical so that we don't get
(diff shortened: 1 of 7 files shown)


src/PAModel/PAConvert/PAWriterVisitor.cs
```diff
@@ -47,13 +47,14 @@ public override void Visit(BlockNode node, Context context)
             node.Name.Accept(this, context);
             context._yaml.WriteStartObject(context._sb.ToString());
 
-            
-            foreach (var func in node.Functions)
+            // Ensure the order we write properties is canonical so that we don't get
+            // noisy diffs on multiple exports. 
+            foreach (var func in node.Functions.OrderBy(x => x.Identifier))
```
MikeStall, line 52: I wanted to enforce ordering in the writer, not the IR, because the IR could be highly mutable. Just do ordering once when we write.

lesaltzm (approved):

Issue #21, opened 2020-11-03
Server side changes to improve .msapp format.


MikeStall:
Improve the real .msapp format to make conversions easier. 

As general principles:
- avoid double json encoding because that breaks checksums. 
- separate out volatile information (timestamps, per-environment) from stable information. 
- remove fields that are redundant or don't need to be emitted. 

Some specific things:

- [ ] Double encoding in PCF controls 
- [ ] Double encoding in Properties.json
- [x] Test cases. (see updates in https://github.com/microsoft/PowerApps-Language-Tooling/issues/52)

MikeStall:
#161 - remove timestamp in sarif files.

lesaltzm:
Sample Resources and Sample Data should not be written to the .msapp. The studio should be able to recreate them on load.

PR #180, opened 2021-03-09
New checksum.

MikeStall:
Better variance and avoids a bunch of collisions.

### DIFF
--- a/src/PAModel/Checksum/IHashMaker.cs
+++ b/src/PAModel/Checksum/IHashMaker.cs
@@ -20,7 +20,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
     {
         void AppendStartObj();
         void AppendPropName(string name);
-        void AppendPropNameSkip(string name);
         void AppendEndObj();
         void AppendStartArray();
         void AppendEndArray();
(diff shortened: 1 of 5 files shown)


src/PAModel/Checksum/IHashMaker.cs
```diff
@@ -39,57 +38,77 @@ class Sha256HashMaker : IHashMaker, IDisposable
...
 
+        private readonly static byte[] _startObj = new byte[] { (byte)'{' };
+        private readonly static byte[] _endObj = new byte[] { (byte) '}' };
+        private readonly static byte[] _startArray = new byte[] { (byte) '[' };
+        private readonly static byte[] _endArray = new byte[] { (byte) ']' };
+        private readonly static byte[] _null = new byte[] { 254 };
+        private readonly static byte[] _true = new byte[] { 1 };
+        private readonly static byte[] _false  = new byte[] { 0 };
+
+        private readonly static byte[] _marker = new byte[] { 255 };
```
MikeStall, line 49: This is the magic - we're making the checksum much more rigid now.

src/PAModelTests/ChecksumTests.cs
```diff
@@ -57,44 +58,54 @@ public void ChecksumsUnique()
...
+        const string C20 = "C6_2YUFpLVLEYtdFvV9iLN8F6TM+cWczemMx4m0VEIpfrg=";
 
 
         [DataTestMethod]
         [DataRow(C1, @"' ab\r\ncd'")] // whitespace
         [DataRow(C1, @"'ab\ncd'")] // same
         [DataRow(C2, @"'ab\ncd  '")] // Trailing whitespace is not trimmed. 
         [DataRow(C3, "{'InvariantScript' : 'abc'}")]
-        [DataRow(C3, "[ 'abc']")] // BAD: InvariantScript is skipped
+        [DataRow(C12, "[ 'abc']")] // InvariantScript is skipped
```
MikeStall, line 90: The thing to notice here is that we remove a bunch of collisions now (values used to be the same, and now they're different).

src/PAModel/Entropy.cs
```diff
@@ -56,6 +56,11 @@ internal class PropertyEntropy
 
         public PropertyEntropy VolatileProperties { get; set; }
 
+        // Sometimes, an empty LocalDatabaseReferences is serialized as "" or { }.
+        // Need to track which so we can round-trip.
+        // if true: "".   Else,  "{}". 
+        public bool LocalDatabaseReferencesAsEmpty { get; set; }
```
lesaltzm, line 62: Which is the more up-to-date style of serializing this? Is there something we should add to #21 to track this?

lesaltzm, line 62: Also, should this be nullable, for when the LocalDatabaseReference isn't empty at all?

lesaltzm (approved):

PR #182, opened 2021-03-10
Write property values in .fx.yaml with platform-appropriate newlines

### LINKED ISSUE #181: Mixed CR and CRLF in Canvas App *.fx.* files after unpack
My unpacked canvas app contains mixes between CR and CRLF after the export:

image

I guess this should be unified.

lesaltzm:
fixes #181

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlWriter.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlWriter.cs
@@ -62,9 +62,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
                 return;
             }
 
-            // Normalize newlines in value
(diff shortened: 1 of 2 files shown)


MikeStall (approved):

src/PAModel/PAConvert/Yaml/YamlWriter.cs
```diff
@@ -61,6 +62,9 @@ public void WriteProperty(string propertyName, string value)
                 return;
             }
 
+            // Normalize newlines in value
+            value = value.Replace("\r\n", "\n").Replace("\r", "\n");
```
MikeStall, line 66: Is there a test we can add for this? 
This is significant change - means we don't round trip exactly, although our checksum is probably flexible enough.

MikeStall (approved):

PR #183, opened 2021-03-10
Only load theme that matches the provided theme name

### LINKED ISSUE #160: How to handle multiple Themes in the Themes.json file
I hope you guys can help me with a custom Theme question.

When I unpack a msapp file, the `Src\Themes.json` file is one of the files that appears.
In an unmodified app, this file contains the 'defaultTheme' and looks like this:

image

Now I want to add my own custom theme, named 'myTheme'. So I add a copy of the defaultTheme to the CustomThemes array, rename it, set the `"CurrentTheme": "myTheme"` and modify the theme with my color modifications etc.
After that, the file looks like this:

image

I can -pack it, open the app in Power Apps Studio and see the Theme changes. All great! But...
When I download the App again (via F12) and -unpack it, the unpacking fails and trips because it finds the `Fill` property twice on the 'defaultScreenStyle' style in the `styles` array.

image

Maybe I do something wrong here, but I would expect that multiple Themes should not conflict with each other, only when the Themes have the same name...

lesaltzm:
resolves #160

### DIFF
--- a/src/PAModel/Serializers/SourceSerializer.cs
+++ b/src/PAModel/Serializers/SourceSerializer.cs
@@ -38,8 +38,9 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         // 16 - Group Control transform
         // 17 - Moved PublishOrderIndex entirely to Entropy 
         // 18 - AppChecker result is not part of entropy (See change 0.5 in this list) 
-        // 19 - Switch extension to .fx.yaml 
-        public static Version CurrentSourceVersion = new Version(0, 19);
+        // 19 - Switch extension to .fx.yaml  
+        // 20 - Only load themes that match the specified theme name 
+        public static Version CurrentSourceVersion = new Version(0, 20);
 
         // Layout is:
         //  src\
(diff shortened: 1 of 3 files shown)


src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -38,8 +38,9 @@ internal static partial class SourceSerializer
         // 16 - Group Control transform
         // 17 - Moved PublishOrderIndex entirely to Entropy 
         // 18 - AppChecker result is not part of entropy (See change 0.5 in this list) 
-        // 19 - Switch extension to .fx.yaml 
-        public static Version CurrentSourceVersion = new Version(0, 19);
+        // 19 - Switch extension to .fx.yaml  
+        // 20 - Only load themes that match the specified theme name 
+        public static Version CurrentSourceVersion = new Version(0, 20);
```
MikeStall, line 43: Is this actually a breaking change? It looks like just a change to the Load path, but not the save path.

lesaltzm: It is, yeah. An app written to source in v19 that gets loaded in v20, if it was using a theme other than the default, won't have the property values that were considered default re-added in the same way.

MikeStall (approved):

PR #190, opened 2021-03-19
Manual layout group container variant doesn't add layout properties

lesaltzm:


src/PAModel/ControlTemplates/DynamicProperties.cs
```diff
@@ -9,9 +9,9 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.ControlTemplates
     // Responsible for handling dynamic properties and their default vaues
     internal static class DynamicProperties
     {
-        internal static bool AddsChildDynamicProperties(string template)
+        internal static bool AddsChildDynamicProperties(string template, string variant)
         {
-            return template == "groupContainer";
+            return template == "groupContainer" && variant != "manualLayoutContainer";
```
MikeStall, line 14: Can we make this a const field? Easy to identify, comment, etc.  Burying the string can be brittle.

MikeStall (approved):

PR #199, opened 2021-04-21
Rewrite asset file paths to use stable meaningful names

### LINKED ISSUE #198: For Asset filenames, use name ("MyImage.svg") instead of number (0003.svg)
Similar to what we did with control filenames to use control name instead of oid. 
(this can be done in a non-breaking way)

lesaltzm:
fixes #198


lesaltzm:
Bump the source version

src/PAModel/Utility/Utilities.cs
```diff
@@ -328,5 +329,25 @@ public static void EnsurePathRooted(string path)
                 throw new ArgumentException("Path must be a full path: " + path);
             }
         }
+
+        /// <summary>
+        /// Relative path of the file depends on the fileType. eg: Image files go into "Assets\Images\"
```
MikeStall, line 334: Can we infer this by file extension instead? (different directories is an unusual pattern for pro-devs)

src/PAModel/CanvasDocument.cs
```diff
@@ -422,8 +426,85 @@ internal HashSet<string> GetImportedComponents()
...
+
+
+        private void StabilizeAssetFilePaths()
+        {
+            _entropy.LocalResourceFileNames.Clear();
+
+            // Update AssetFile paths
+            foreach (var resource in _resourcesJson.Resources)
+            {
+                if (resource.ResourceKind != "LocalFile")
```
MikeStall, line 438: nit: can we put this string literal as a constant? als omakes it easier to comment it.

src/PAModel/CanvasDocument.cs
```diff
@@ -422,8 +426,85 @@ internal HashSet<string> GetImportedComponents()
...
+                    continue;
+
+                var assetFilePath = FilePath.FromMsAppPath(resource.Path.Substring(AssetFilePathPrefix.Length));
```
MikeStall, line 441: nit: can we wrap this part in a  function? We repeat it 4x, the substring on file path is brittle. 
```
FilePath.FromMsAppPath(resource.Path.Substring(AssetFilePathPrefix.Length));
```

src/PAModel/Utility/Utilities.cs
lesaltzm: Yes, absolutely. We'll still need this function for the msapp side, but we could dump them flat in Assets/

src/PAModel/CanvasDocument.cs
lesaltzm: yup

src/PAModel/Utility/Utilities.cs
lesaltzm: Discussed offline, keeping.

MikeStall (approved):

PR #203, opened 2021-04-28
Fixes an issue with directories that contain a period (.)

### LINKED ISSUE #202: Issue using the `pack` command when the directory being packed includes a period (.) in its name
The logic in the `Utilities.AppendDirectorySeparatorChar` [here](https://github.com/microsoft/PowerApps-Language-Tooling/blob/c443fc7823c25d91e9f653aef6236095299382f5/src/PAModel/Utility/Utilities.cs#L173-L183) does not consider directories that include a period (.) in their names, ie "C:\SomePath\SomeSolution\SomeSolution.SomeProject"

Since the `Path.HasExtension` method returns `true` in this instance, the `Path.DirectorySeparatorChar` will not be appended.

StevenRasmussen:
Fixes #202 

I've had this code running in an app of mine successfully for some time.

### DIFF
--- a/src/PAModel/Utility/Utilities.cs
+++ b/src/PAModel/Utility/Utilities.cs
@@ -36,11 +36,11 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             return list;
         }
 
-        public static void AddRange<TKey,TValue>(
+        public static void AddRange<TKey, TValue>(
             this IDictionary<TKey, TValue> thisDictionary,
             IEnumerable<KeyValuePair<TKey, TValue>> other)
         {
-            foreach(var kv in other)
+            foreach (var kv in other)
             {
                 thisDictionary[kv.Key] = kv.Value;
             }
@@ -57,7 +57,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             return @default;
         }
 
-        public static TValue GetOrCreate<TKey,TValue>(this IDictionary<TKey, TValue> dict, TKey key)
+        public static TValue GetOrCreate<TKey, TValue>(this IDictionary<TKey, TValue> dict, TKey key)
             where TValue : new()
         {
             TValue value;
@@ -173,8 +173,20 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         private static string AppendDirectorySeparatorChar(string path)
         {
             // Append a slash only if the path is a directory and does not have a slash.
-            if (!Path.HasExtension(path) &&
-                !path.EndsWith(Path.DirectorySeparatorChar.ToString()))
+            var isFilePath = File.Exists(path);
+            var isDirectoryPath = Directory.Exists(path);
+
+            var pathExists = isFilePath || isDirectoryPath;
+            if (pathExists && isFilePath)
+            {
+                return path;
+            }
+            else if (pathExists && isDirectoryPath)
+            {
+                if (!path.EndsWith(Path.DirectorySeparatorChar.ToString()))
+                    return path + Path.DirectorySeparatorChar;
+            }
+            else if (!Path.HasExtension(path) && !path.EndsWith(Path.DirectorySeparatorChar.ToString()))
             {
                 return path + Path.DirectorySeparatorChar;
             }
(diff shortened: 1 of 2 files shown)


src/PAModel/Utility/Utilities.cs
```diff
@@ -173,8 +173,20 @@ public static string GetRelativePath(string full, string basePath)
         private static string AppendDirectorySeparatorChar(string path)
```
MikeStall, line 173: Thanks for looking at this. 

Can you add a test case to  UtilityTests.cs!TestRelativePath to cover this case?

src/PAModelTests/UtilityTests.cs
```diff
@@ -36,29 +37,35 @@ public void TestNotEscaped()
...
-        // [DataRow("C:\\Foo\\Bar\\Baz", "C:\\Foo\\", "Bar\\Baz\\")]
-        // [DataRow("C:\\Foo\\Bar\\Baz\\", "C:\\Foo\\", "Bar\\Baz\\")]
-        // [DataRow("C:\\Foo\\Bar.msapp", "C:\\Foo", "Bar.msapp")]
-        // [DataRow("C:\\Foo\\Bar.msapp", "C:\\Foo\\", "Bar.msapp")]
-        // [DataRow("C:\\Foo\\Bar.msapp", "C:\\", "Foo\\Bar.msapp")]
-        [DataRow(@"C:\DataSources\JourneyPlanner|Sendforapproval.json", "C:\\", @"DataSources\JourneyPlanner|Sendforapproval.json")]
-        [DataRow(@"C:\DataSources\JourneyPlanner%7cSendforapproval.json", "C:\\", @"DataSources\JourneyPlanner%7cSendforapproval.json")]
-        // [DataRow(@"d:\app\Src\EditorState\Screen%252.editorstate.json",
-        //     @"d:\app", @"Src\EditorState\Screen%252.editorstate.json" )]
+        [DataRow(@"C:\Temp\DataSources\JourneyPlannerSendforapproval.json", @"C:\Temp", @"DataSources\JourneyPlannerSendforapproval.json")]
```
StevenRasmussen, line 40: I added the sub directory `Temp` to the path since this file will get written to the hard drive now.  I figured putting it in a folder called `Temp` would be better than having a directory called `DataSources` on the root of the hard drive. Will this cause any issues with an automated build/test?

src/PAModel/Utility/Utilities.cs
StevenRasmussen: Added the test case as requested.

src/PAModel/Utility/Utilities.cs
```diff
@@ -173,8 +173,20 @@ public static string GetRelativePath(string full, string basePath)
...
         {
             // Append a slash only if the path is a directory and does not have a slash.
-            if (!Path.HasExtension(path) &&
-                !path.EndsWith(Path.DirectorySeparatorChar.ToString()))
+            var isFilePath = File.Exists(path);
+            var isDirectoryPath = Directory.Exists(path);
```
MikeStall, line 177: This should be a "pure" string operation, we shouldn't depend on the file system.  (I realized this after looking at the tests). 
We'd like to just call Path.GetRelativePath() - that correctly handles the inputs here -  but it's not available in this framework, so we handrolled it ourselves. Interesting to see where they added it:   https://github.com/dotnet/corefx/pull/11689/files

Their impl is significantly more complex than the impl here, which suggests that there are other corner cases we're missing.

StevenRasmussen: That's certainly ideal ;)  Based on the `Path.GetRelativePath` implementation it states that the `relativeTo` parameter is ALWAYS considered to be a directory.  If we also make that the premise for the custom implementation here then the problem becomes much simpler.  It would also make transitioning to the NetCore version down the road pretty seamless.  Thoughts?

MikeStall: The current impl is at:  https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L861

That implementation is pretty complex (lots of helpers, platform-specific code, unsafe pointers), so it looks like we can't just port it. 

The 'relativeTo' parameter is (basePath) is always a directory.

MikeStall: Actually, 1st arg is always file, 2nd arg is always dir, so that should simplify this. 
Here's a proposal: https://github.com/microsoft/PowerApps-Language-Tooling/pull/204 

How's that look? (github won't let me add you as a reviewer).

Some of the tests were accidentally commented out, so I reenabled those. 
Plus one test was bogus.

StevenRasmussen:
Cancel in lieu of #204

PR #206, opened 2021-05-04
Fix asset file name collision

lesaltzm:


src/PAModel/Schemas/ResourcesJson.cs
```diff
@@ -42,6 +42,9 @@ internal class ResourceJson
         // We preserve it in Entropy.json for roundtrip purposes only.
         public string RootPath { get; set; }
 
+        // Original filename used if there were collisions when rewriting the filenames
+        public string OriginalName { get; set; }
```
MikeStall, line 46: This field isn't present on server? Just used for client representation?

src/PAModel/CanvasDocument.cs
```diff
@@ -445,6 +445,20 @@ private void StabilizeAssetFilePaths()
                 if (!_assetFiles.TryGetValue(assetFilePath, out var fileEntry))
                     continue;
 
+                if (_entropy.LocalResourceFileNames.ContainsKey(resource.Name))
```
MikeStall, line 448: The order of items in the resources array in Resources.json appears to be random. 
Which means our traversal here could be random. 
Which means we may randomly decide which resource gets the "image.jpg" and which faults over to "image_1".jpg. 
Which could lead to noisy diffs. 

Probably easiest fix is to sort the traversal on line 439.

MikeStall:
Filed #207 to track the bug here.

src/PAModel/CanvasDocument.cs
```diff
@@ -445,6 +445,20 @@ private void StabilizeAssetFilePaths()
...
+                {
+                    int i = 1;
+                    var newResourceName = resource.Name + '_' + i;
```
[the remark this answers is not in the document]
suskumar-MSFT: I have the same concern, it might confuse the user as well which one is actual image_1?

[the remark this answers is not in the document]
lesaltzm: As discussed offline, we think that this is probably ok, but added a warning message to make it clearer for end users when this occurs. 
image

src/PAModel/CanvasDocument.cs
```diff
@@ -445,6 +445,23 @@ private void StabilizeAssetFilePaths()
...
+                    while (_entropy.LocalResourceFileNames.ContainsKey(newResourceName))
```
suskumar-MSFT, line 453: @lesaltzm will this handle that scenario elegantly when the after renaming the new name becomes image_1.jpg but file with image_1.jpg exists already. In that case it should go another round of renaming which should make the newName = image_1_1.jpg

lesaltzm: Right, so, in that case, if we had `image`, `Image`, and `image_1`, we wind up with `image` `Image_1` and `image_1_1`

It's not elegant, per-se. We could do this as two passes, where we first populate the names, and then do renames, that might be better.

suskumar-MSFT: ok was just checking if it handles this, and not try to persist two files with image_1 names which could break unpacking

lesaltzm: Ok, pushed a better rename, with test. 
Now we get 
`image`, `Image_2` and `image_1`

src/PAModel/CanvasDocument.cs
```diff
@@ -499,6 +537,12 @@ private void RestoreAssetFilePaths()
...
 
+                if (resource.OriginalName != null)
+                {
+                    resource.Name = resource.OriginalName;
+                }
+
+                resource.OriginalName = null;
                 var updatedPath = FilePath.FromMsAppPath(Utilities.GetResourceRelativePath(resource.Content)).Append(msappFileName);
                 resource.Path = updatedPath.ToMsAppPath();
                 resource.FileName = msappFileName;
```
MikeStall, line 548: Are the mutations here safe? 
We get here via: SaveAsMsapp() --> ApplyBeforeMsAppWriteTransforms() --> RestoreAssetFilePaths(). 

Ie, does this cause a mutation of the app._resource after we call save?

src/PAModel/CanvasDocument.cs
```diff
@@ -431,20 +431,58 @@ private FilePath GetAssetFilePathWithoutPrefix(string path)
...
+            var caseInsensitiveNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var caseSensitiveNames = new HashSet<string>(StringComparer.Ordinal);
+            foreach (var resource in _resourcesJson.Resources.OrderBy(resource => resource.Name, StringComparer.Ordinal))
+            {
+                if (resource.ResourceKind != ResourceKind.LocalFile)
+                    continue;
+
+                if (!caseInsensitiveNames.Contains(resource.Name))
+                {
+                    caseInsensitiveNames.Add(resource.Name);
```
MikeStall, line 449: nit: Add() returns a boolean, so I think this could be: 

if (caseInsensitiveNames.Add(resource.Name))
{
 caseSensitiveNames.Add(resource.Name);
}

src/PAModel/CanvasDocument.cs
lesaltzm: basically yes it's safe it's guarded the same way the other transforms are

MikeStall (approved):

PR #208, opened 2021-05-05
Changes to abstract out InstumentationKey from Properties.json to a separate file

suskumar-MSFT:
Changes to abstract out InstumentationKey from Porperties.json to a separate file - ApplicationInsightsKey.json

### DIFF
--- a/src/PAModel/Schemas/adhoc/AppInsightsKey.cs
+++ b/src/PAModel/Schemas/adhoc/AppInsightsKey.cs
@@ -1,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.PowerPlatform.Formulas.Tools
+{
+    // Abstract out the InstrumentationKey (From AppInsights) of an app in a seprate json file (AppInsightsKey.json)
+    class AppInsightsKeyJson
+    {
+        public string InstrumentationKey { get; set; } // a guid
+
+        [JsonExtensionData]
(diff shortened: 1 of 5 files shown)


src/PAModel/Schemas/adhoc/AppInsightsKey.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.PowerPlatform.Formulas.Tools
+{
+    // Abstract out the InstrumentationKey (From AppInsights) of an app in a seprate json file (AppInsightsKey.json)
+    class AppInsightsKeyJson
+    {
+        public string InstrumentationKey { get; set; } // a guid
+
+        [JsonExtensionData]
+        public Dictionary<string, JsonElement> ExtensionData { get; set; }
```
MikeStall, line 18: I think we can skip this - this json file is only in the extract case, so we control all the fields here.

MikeStall: put another way, any extra fields here would *not* be added into the msapp.

suskumar-MSFT: yea I was thinking the same, will remove it.

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -274,6 +274,11 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
                     app._properties.LocalDatabaseReferences = null;
                     app._entropy.LocalDatabaseReferencesAsEmpty = false;
                 }
+                if (!string.IsNullOrEmpty(app._properties.InstrumentationKey))
```
MikeStall, line 277: When we roundtrip, we'll need to preserve between null and "".

MikeStall (approved):

PR #210, opened 2021-05-07
Fixing round trip errors for Command Component

suskumar-MSFT:
Currently the tool fails to de-serialize the component library defined for Command components, adding support for same.


MikeStall (approved):
:shipit:

PR #214, opened 2021-05-12
Handling for empty LocalDatabaseReferences which has valid CDS datasource

suskumar-MSFT:
Made the changes to handle for empty LocalDatabaseReferences property in proerties.json when the app has default cds as data source. This is happening for command components.

Also in certain cases the data source in table definition is null which causes failure during round trip, added fix for that as well.

### DIFF
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -253,14 +253,14 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 // Normalize logo filename. 
                 app.TranformLogoOnLoad();
 
-                if (app._properties.LibraryDependencies != null)
+                if (!string.IsNullOrEmpty(app._properties.LibraryDependencies))
                 {
                     var refs = Utilities.JsonParse<ComponentDependencyInfo[]>(app._properties.LibraryDependencies);
                     app._libraryReferences = refs;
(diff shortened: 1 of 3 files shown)


lesaltzm (approved):

MikeStall (approved):
:shipit:

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -253,14 +253,14 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
                 // Normalize logo filename. 
                 app.TranformLogoOnLoad();
 
-                if (app._properties.LibraryDependencies != null)
+                if (!string.IsNullOrEmpty(app._properties.LibraryDependencies))
```
MikeStall, line 256: >string.IsNullOrEmpty [](http://example.com/codeflow?start=21&length=20)

The one danger to be aware of here is that this won't round-trip: we'll read in empty string and write out null. 
That will result in a checksum mismatch. 
But that's probalby fine in this case (since this is just extra hardening and the server should alway write out null).

suskumar-MSFT: the problem is if we don't check for empty string the code would throw an exception in the very next line where it would try to deserialize empty string as json, yes we can probably do special handling for empty strings vs null but haven't seen any such instance yet.

PR #219, opened 2021-05-14
Ensure group controls with missing editor state load correctly in Studio

### LINKED ISSUE #209: The groupcontroltransform pass does not work correctly without editorstate

lesaltzm:
Fixes #209 
In the control .json file, group controls must be ordered after the controls they are grouping of, to ensure that they are reparented correctly when loading.


MikeStall (approved):
Is there some test we can add here? (Whether a unit test, or catch it in a round-tripping test).

lesaltzm:
> Is there some test we can add here? (Whether a unit test, or catch it in a round-tripping test).

Added a test

suskumar-MSFT (approved):

Issue #215, opened 2021-05-13
Show error if a pkg xml is missing 


MikeStall:
On pack, show a warning if a yaml file references a control template (ie, pkgs\label_2.3.1.xml) that's missing. 

Bonus if we can give a good error reference a location in the yaml that Refers to the pkg.

suskumar-MSFT:
#220

PR #220, opened 2021-05-18
#215: Adding validation to check for the existence of the templates 

suskumar-MSFT:
Adding validation to check for the existence of the templates which are referred in the sources.
* Added validation for widget templates
* Added validation for component templates
* Added validation for PCF templates

The validation would not fail for widget templates for existing unpacked sources.

### DIFF
--- a/src/PAModel/EditorState/CombinedTemplateState.cs
+++ b/src/PAModel/EditorState/CombinedTemplateState.cs
@@ -13,8 +13,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.EditorState
     // A combination of the control templates present in Templates.json and the control files
     internal class CombinedTemplateState
     {
-        private bool _isXmlTemplate = false;
-
         public string Id { get; set; }
 
         // Very important for data components.
@@ -52,17 +50,8 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.EditorState
         // Maps to TemplateMetadataJson ExtensionData
         public Dictionary<string, object> ComponentExtraMetadata { get; set; }
 
-        public bool IsWidgetTemplate
-        {
-            get
-            {
-                return _isXmlTemplate;
-            }
-            set
-            {
-                _isXmlTemplate = value;
-            }
-        }
+        // Some of the xml templates have widget root node, and these are templates which are retruned by the doc server in the UsedTemplates collection.
+        public bool IsWidgetTemplate { get; set; }
 
 
         public CombinedTemplateState() { }
--- a/src/PAModel/EditorState/CombinedTemplateState.cs
+++ b/src/PAModel/EditorState/CombinedTemplateState.cs
@@ -13,6 +13,8 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.EditorState
     // A combination of the control templates present in Templates.json and the control files
     internal class CombinedTemplateState
     {
+        private bool _isXmlTemplate = false;
+
         public string Id { get; set; }
 
         // Very important for data components.
@@ -49,6 +51,18 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.EditorState
 
         // Maps to TemplateMetadataJson ExtensionData
         public Dictionary<string, object> ComponentExtraMetadata { get; set; }
+
+        public bool IsWidgetTemplate
+        {
+            get
+            {
+                return _isXmlTemplate;
+            }
+            set
+            {
+                _isXmlTemplate = value;
+            }
+        }
 
 
         public CombinedTemplateState() { }
--- a/src/PAModel/EditorState/CombinedTemplateState.cs
+++ b/src/PAModel/EditorState/CombinedTemplateState.cs
@@ -50,7 +50,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.EditorState
         // Maps to TemplateMetadataJson ExtensionData
         public Dictionary<string, object> ComponentExtraMetadata { get; set; }
 
-        // Some of the xml templates have widget root node, and these are templates which are retruned by the doc server in the UsedTemplates collection.
+        // Some of the xml templates have widget root node, and these are templates which are returned by the doc server in the UsedTemplates collection.
         public bool IsWidgetTemplate { get; set; }
 
 
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -195,6 +195,14 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                     }
                 } // foreach zip entry
 
+                foreach(var template in app._templates.UsedTemplates)
+                {
+                    if(app._templateStore.TryGetTemplate(template.Name, out var templateState))
+                    {
+                        templateState.IsWidgetTemplate = true;
+                    }
+                }
+
                 foreach (var componentTemplate in app._templates.ComponentTemplates ?? Enumerable.Empty<TemplateMetadataJson>())
                 {
                     if (!app._templateStore.TryGetTemplate(componentTemplate.Name, out var template))
--- a/src/PAModel/Serializers/SourceSerializer.cs
+++ b/src/PAModel/Serializers/SourceSerializer.cs
@@ -58,7 +58,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         public static readonly string SwaggerPackageDir = Path.Combine("pkgs", "Swagger");
         public static readonly string ComponentPackageDir = Path.Combine("pkgs", "Components");
         public const string OtherDir = "Other";
-        public const string EntropyDir = "Entropy";  
+        public const string EntropyDir = "Entropy";
         public const string ConnectionDir = "Connections";
         public const string DataSourcesDir = "DataSources";
 
@@ -124,7 +124,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                         break;
                 }
             }
-            if(appInsightsInstumentationKey != null)
+            if (appInsightsInstumentationKey != null)
             {
                 app._properties.InstrumentationKey = appInsightsInstumentationKey;
             }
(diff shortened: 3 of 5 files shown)


src/PAModel/EditorState/CombinedTemplateState.cs
```diff
@@ -50,6 +52,18 @@ internal class CombinedTemplateState
         // Maps to TemplateMetadataJson ExtensionData
         public Dictionary<string, object> ComponentExtraMetadata { get; set; }
 
+        public bool IsWidgetTemplate
```
MikeStall, line 55: Can you add a comment about what this means?

nit:
IsWidgetTemplate {get;set;} 
and skip the field.

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -195,6 +195,14 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
                     }
                 } // foreach zip entry
 
+                foreach(var template in app._templates.UsedTemplates)
```
lesaltzm, line 198: nit: Run your editor's code formatter before pushing. 
(space after `foreach (var`)

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +797,78 @@ private static BuildVerJson GetBuildDetails()
             }
         }
 
+        private static void ValidateIfTemplateExists(CanvasDocument app, BlockNode node, ErrorContainer errors)
+        {
+            // parse all children of the root nodes to see if al of them have the temeplate they require.
```
MikeStall, line 802: >al of them have the temeplate [](http://example.com/codeflow?start=62&length=29)

nit: typo

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +797,78 @@ private static BuildVerJson GetBuildDetails()
...
+            foreach(var child in node.Children)
+            {
+                if (string.IsNullOrEmpty(child?.Name?.Kind?.TypeName))
```
lesaltzm, line 805: under what circumstances are child, name, kind, or typename null? As far as I can tell, they should all always have values

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +797,78 @@ private static BuildVerJson GetBuildDetails()
...
+                }
+                return;
+            }
+            
+            CombinedTemplateState templateState;
+            app._templateStore.TryGetTemplate(childNode.Name.Kind.TypeName, out templateState);
+            // For each of the template there mus be an entry in the _templateStore.
+            if(templateState == null)
+            {
+                errors.ValidationError($"Could not find the template with name: {childNode.Name.Kind.TypeName} which is referred in {parentNode.Name.Identifier}. " +
```
MikeStall, line 849: >referred in {parentNode.Name.Identifier}. [](http://example.com/codeflow?start=120&length=41)

can we pass in the source location? (IrNode.SourceSpan). 
ValidateError() should have another overload that accepts source loc.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +797,78 @@ private static BuildVerJson GetBuildDetails()
             }
         }
 
+        private static void ValidateIfTemplateExists(CanvasDocument app, BlockNode node, ErrorContainer errors)
```
lesaltzm, line 800: What is the purpose of this wrapper? it looks like the start of ValidateIfTemplateExistsInternal is basically the same?

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +797,78 @@ private static BuildVerJson GetBuildDetails()
...
+
+        private static void ValidateIfTemplateExistsInternal(CanvasDocument app, BlockNode parentNode, BlockNode childNode, ErrorContainer errors)
+        {
+            if (string.IsNullOrEmpty(childNode?.Name?.Kind?.TypeName))
+            {
+                // could not perform validation.
+                return;
+            }
+            // group and grouContainer controls aren't real controls they contain other child controls.
+            if (childNode.Name.Kind.TypeName == "group" || childNode.Name.Kind.TypeName == "groupContainer")
```
lesaltzm, line 835: What about galleries and other nested controls? We still need to validate the child controls for them as well.

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: they should never be null I haven't had any instance of this, just a safe check.

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: this wrapper takes in node from one of the source files, basically the parent root node like a screen. This wrapper is iterating through the children of the screen and the the other methods makes sure to iterate through all the other individual controls. 
I was trying to do it in one method recursively but found some instances where non group/groupContainer controls also had children.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +797,78 @@ private static BuildVerJson GetBuildDetails()
...
+            CombinedTemplateState templateState;
+            app._templateStore.TryGetTemplate(childNode.Name.Kind.TypeName, out templateState);
+            // For each of the template there mus be an entry in the _templateStore.
+            if(templateState == null)
+            {
+                errors.ValidationError($"Could not find the template with name: {childNode.Name.Kind.TypeName} which is referred in {parentNode.Name.Identifier}. " +
+                    $"If the template was deleted intentionally please make sure to update the source files to remove the references to this template.");
+            }
+            // If its a widget template then there must be a xml file in the pkgs directory.
+            if(templateState.IsWidgetTemplate && !app._templates.UsedTemplates.Any(x => x.Name.Equals(childNode.Name.Kind.TypeName, StringComparison.InvariantCultureIgnoreCase)))
```
MikeStall, line 853: >InvariantCultureIgnoreCase [](http://example.com/codeflow?start=149&length=26)

double checking string compares here -  why is this one IgnoreCase, but the one on line 859 isn't?

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: can I assume that if any control has any children run the validation for all the children, so that I don't have to check for certain specific controls? Currently I was only running the validation on children for group/groupContainer controls, probably I can run for all controls with children.

lesaltzm: yup, that's a good approach.

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: have removed these checks since this should never happen, and if it happens there is something wrong and our tool should fail.

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: Removed these ignore case checks, doesn't seem like these are needed.

src/PAModel/EditorState/CombinedTemplateState.cs
```diff
@@ -50,6 +50,9 @@ internal class CombinedTemplateState
         // Maps to TemplateMetadataJson ExtensionData
         public Dictionary<string, object> ComponentExtraMetadata { get; set; }
 
+        // Some of the xml templates have widget root node, and these are templates which are retruned by the doc server in the UsedTemplates collection.
```
MikeStall, line 53: >retruned [](http://example.com/codeflow?start=94&length=8)

typo

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +798,62 @@ private static BuildVerJson GetBuildDetails()
...
+                {
+                    continue;
+                }
+
+                // If its a widget template then there must be a xml file in the pkgs directory.
+                if (templateState.IsWidgetTemplate)
+                {
+                    if (!app._templates.UsedTemplates.Any(x => x.Name == child.Name.Kind.TypeName))
+                    {
+                        errors.ValidationError(root.SourceSpan.GetValueOrDefault(), $"Widget control template: {templateState.Name}, version {templateState.Version} was not found in the pkgs directory and is referred in {root.Name.Identifier}. " +
```
MikeStall, line 831: >root [](http://example.com/codeflow?start=47&length=4)

is the error location here child or root?

MikeStall, line 831: >.GetValueOrDefault() [](http://example.com/codeflow?start=62&length=20)

Can omit this.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -795,5 +798,62 @@ private static BuildVerJson GetBuildDetails()
...
+                    {
+                        errors.ValidationError(root.SourceSpan.GetValueOrDefault(), $"Component template: {templateState.Name} was not found in Src/Components directory and is referred in {root.Name.Identifier}. " +
+                            $"If the template was deleted intentionally please make sure to update the source files to remove the references to this template.");
+                    }
+                    continue;
+                }
+                // PCF are dynamically imported controls and their template definition is stored in the DynamicControlDefinitionJson property, check if that exists.
+                else if (templateState.Id.StartsWith("http://microsoft.com/appmagic/powercontrol"))
+                {
+                    if (!templateState.ExtensionData.ContainsKey("DynamicControlDefinitionJson") || templateState.ExtensionData["DynamicControlDefinitionJson"] == null)
```
MikeStall, line 849: >DynamicControlDefinitionJson [](http://example.com/codeflow?start=66&length=28)

maintenance nit: can we use a nameof() expression here
(in general, we should avoid string constants like this)

MikeStall (approved):
:shipit:

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: root is the original source which is being processed, the child could be groupControl, Gallery or any control which can have nested controls. so I wanted to point which source file have these issues.

suskumar-MSFT: the AddError takes in "SourceLocation" where the root.SourceSpan is of nullable type "SoruceLocation?", so either I can do the default handling here (like its currently done) or can do that in the ValidateError overload. This seems to be a better palce to handle.

Issue #211, opened 2021-05-10
"Not supported" should be its own error, easy to filter out. 


MikeStall:
There are cases the tool doesn't support (like old CDS connectors). 
We want to filter out "not supported" from "tool crashes". 
These should give a dedicated ErrorCode (4001) rather than the internal errors (30xx).

PR #221, opened 2021-05-18
#211: adding new error type for unsupported scenarios.

suskumar-MSFT:
Adding a new error code for scenarios which are not supported by the tool.
Eg: "Connection Accounts is using the old CDS connector which is incompatible with this tool"


MikeStall (approved):
:shipit:

lesaltzm (approved):

Issue #217, opened 2021-05-13
Update Assets/ + Resource.json to enable eliminating or significantly reducing resource.json


lesaltzm:

MikeStall:
fyi, supported file extensions:
```
    export const ImageFormats = ['.jpg', '.jpeg', '.gif', '.png', '.bmp', '.tif', '.tiff', '.svg'];
    export const MediaFormats = ['.wav', '.mp3', '.mp4', '.wma', '.wmv'];
    export const AudioFormats = ['.wav', '.mp3', '.wma'];
    export const VideoFormats = ['.mp4', '.wmv'];

```

PR #223, opened 2021-05-20
#217: Made changes to to not persist LocalFile entries in Resources.json

suskumar-MSFT:
Added changes to not persist the local asset file entries in Resources.json, and then add it back dynamically when the msapp is being packaged again.


src/PAModel/Schemas/ResourcesJson.cs
```diff
@@ -24,6 +24,13 @@ internal enum ResourceKind
         Uri,
     }
 
+    internal enum Schema
+    {
+        i,
+        m,
+        o
```
MikeStall, line 31: Can you comment these?

src/PAModel/Entropy.cs
```diff
@@ -43,6 +43,9 @@ internal class PropertyEntropy
         // Key is component name, value is Index. 
         public Dictionary<string, double> ComponentIndexes { get; set; } = new Dictionary<string, double>(StringComparer.Ordinal);
 
+        // Key is resource name, value is Index. 
+        public Dictionary<string, double> ResourceJsonIndexes { get; set; } = new Dictionary<string, double>(StringComparer.Ordinal);
```
MikeStall, line 47: >double [](http://example.com/codeflow?start=34&length=6)

Curious - why are these all doubles instead of int?

MikeStall: Actually.. for assets, they have names like "0001.png". Should the entropy capture the string name instead of the index. We try to regenerate a str from index, but that could be brittle (can we gaurantee server always files that format and casing?)

MikeStall:
Be sure this still repacks and loads properly even if entropy is deleted. 
This is something you could test - in one of the round-trip tests, delete entropy, and validate the resulting resources.json. (or potentially even have a 2nd baseline for repacking w/ a deleted entropy)

src/PAModel/Entropy.cs
suskumar-MSFT: let me take a look in the server to see if that's guaranteed, if so then I can get rid of this extra dictionary, since we already capture the mapping of the original names and the new names in LocalResourceFileNames dictionary

suskumar-MSFT: I confirmed this only the LocalFile resources have these kind of names the Uri resources don't follow this convention.

src/PAModel/CanvasDocument.cs
```diff
@@ -81,6 +81,7 @@ public class CanvasDocument
 
         // Track all asset files, key is file name
         internal Dictionary<FilePath, FileEntry> _assetFiles = new Dictionary<FilePath, FileEntry>();
+        internal Dictionary<string, LocalAssetInfoJson> _localAssetInfoJson = new Dictionary<string, LocalAssetInfoJson>();
```
lesaltzm, line 84: Can you add a comment about what the key is, and also specify what comparison is being used (Case sensitive, insensitive, ...)

src/PAModel/Schemas/LocalResourceInfo.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.Formulas.Tools.Schemas
+{
+    internal class LocalAssetInfoJson
```
lesaltzm, line 6: Add a comment calling out that this schema is only  for the unpacked side, not part of the .msapp

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -138,14 +138,75 @@ public static CanvasDocument LoadFromSource(string directory2, ErrorContainer er
...
+            
+            foreach (var file in dir.EnumerateFiles(AssetsDir, "*", false))
+            {
+                var fileEntry = file.ToFileEntry();
+                // although should never be the case, but if it happens that both Resources.json file and individual sharded files exist, then read both.
                 if (file._relativeName == "Resources.json")
                 {
                     app._resourcesJson = file.ToObject<ResourcesJson>();
-                    continue;
                 }
```
lesaltzm, line 152: Are we actually backwards compatible here? If not, we should just remove this bit and bump the source version at the top of this file

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -138,14 +138,75 @@ public static CanvasDocument LoadFromSource(string directory2, ErrorContainer er
...
-                app.AddAssetFile(file.ToFileEntry());
+                else if(fileEntry.Name.GetExtension() == ".json")
+                {
+                    // if its a json file then this must be one of the sharded files from Resources.json
+                    resources.Add(file.ToObject<ResourceJson>());
```
lesaltzm, line 156: Style nit on all of these, `// Capital Letters at the start of a comment`

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -138,14 +138,75 @@ public static CanvasDocument LoadFromSource(string directory2, ErrorContainer er
...
+                }
+                else
+                {
+                    // if its not a json file then add it to the asset files.
+                    // eg. logo file.
+                    app.AddAssetFile(fileEntry);
+                }
+            }
+
+            if(resources.Count > 0)
```
lesaltzm, line 166: nit: run your editor's formatter

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -138,14 +138,75 @@ public static CanvasDocument LoadFromSource(string directory2, ErrorContainer er
...
+            {
+                // if there is no Resources.json file then app._resourcesJson would be null, so make sure to initialize it before adding resource entries.
+                if (app._resourcesJson == null)
+                {
+                    app._resourcesJson = new ResourcesJson();
+                }
+
+                // although we should not have resouces.json file when the individual sharded files are there.
+                // but if that happens then try to do a union of all the resources in Resources.json and the sharded files.
```
lesaltzm, line 175: If we don't expect resources.json, we should not have this logic and instead just bump the version

src/PAModel/Schemas/ResourcesJson.cs
```diff
@@ -36,17 +47,16 @@ internal class ResourceJson
         public ResourceKind ResourceKind { get; set; }
         public string Path { get; set; }
         public ContentKind Content { get; set; }
+        public string Schema { get; set; }
```
lesaltzm, line 50: Does this need to be a string, or can it be the enum? System.Text.Json should be able to handle serializing/deserializing the enum written out

src/PAModel/Serializers/TransformResourceJson.cs
```diff
@@ -0,0 +1,141 @@
...
+                return ContentKind.Audio;
+            else if (VideoExtensionRegEx.IsMatch(name))
+                return ContentKind.Video;
+            else if (PdfExtensionRegEx.IsMatch(name))
+                return ContentKind.Pdf;
+            else
+                return ContentKind.Unknown;
+        }
+
+        private static string GetSchema(ContentKind contentKind)
```
lesaltzm, line 116: If you update the schema to use the enum, this gets simpler.

suskumar-MSFT: the problem is I can't have '?' in an enum, which is the value of schema used by server for unknown types, now sure if we can ever have such type...

src/PAModel/Schemas/ResourcesJson.cs
suskumar-MSFT: Can't use enum type for the the character used for unknown types '?'

MikeStall (approved):

lesaltzm (approved):

Issue #226, opened 2021-05-25
test case: deleting entropy


MikeStall:
Can you add a test case which deletes entropy and repacks?  This may need 2 msapps (the original, and the repacked-without-entropy).  
Similar to what we do in RoundtripTests (that just verifies we can repack exactly).

PR #229, opened 2021-05-27
Changes to ensure the table definitions are added last if the entropy information is missing.

suskumar-MSFT:
Table Definitions need to be added after the ViewInfo elements to ensure serialization order of the server is respected.

### DIFF
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -538,8 +538,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 yield return ToFile(FileKind.PublishInfo, publishInfo);
 
             // "DataComponent" data sources are not part of DataSource.json, and instead in their own file
-            // In case the the entropy is missing then make sure that tableDefinitions are being added towards the end in DataSources.json file
-            // This is to respect the serialization order of the server.
             var dataSources = new DataSourcesJson
             {
                 DataSources = app.GetDataSources()
@@ -548,7 +546,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                     .OrderBy(x => app._entropy.GetOrder(x))
                     .ToArray()
             };
-
             yield return ToFile(FileKind.DataSources, dataSources);
 
             var sourceFiles = new List<SourceFile>();
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -537,14 +537,22 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 yield return ToFile(FileKind.PublishInfo, publishInfo);
 
             // "DataComponent" data sources are not part of DataSource.json, and instead in their own file
+            // In case the the entropy is missing then make sure that tableDefinitions are being added towards the end in DataSources.json file
+            // This is to respect the serialization order of the server.
             var dataSources = new DataSourcesJson
             {
                 DataSources = app.GetDataSources()
                     .SelectMany(x => x.Value)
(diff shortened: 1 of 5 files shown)


src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -537,14 +537,22 @@ private static IEnumerable<FileEntry> GetMsAppFiles(this CanvasDocument app, Err
...
             // "DataComponent" data sources are not part of DataSource.json, and instead in their own file
+            // In case the the entropy is missing then make sure that tableDefinitions are being added towards the end in DataSources.json file
+            // This is to respect the serialization order of the server.
             var dataSources = new DataSourcesJson
             {
                 DataSources = app.GetDataSources()
                     .SelectMany(x => x.Value)
-                    .Where(x => !x.IsDataComponent)
+                    .Where(x => !x.IsDataComponent && x.TableDefinition == null)
                     .OrderBy(x => app._entropy.GetOrder(x))
```
MikeStall, line 547: >GetOrder [](http://example.com/codeflow?start=47&length=8)

Can we just make GetOrder() be "smarter", ie when there's no entropy , it can put TableDefinitions at the end. 

There's a potential issue here: this could potentially reorder from what's in entropy.

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -537,14 +537,22 @@ private static IEnumerable<FileEntry> GetMsAppFiles(this CanvasDocument app, Err
...
 
             // "DataComponent" data sources are not part of DataSource.json, and instead in their own file
+            // In case the the entropy is missing then make sure that tableDefinitions are being added towards the end in DataSources.json file
+            // This is to respect the serialization order of the server.
             var dataSources = new DataSourcesJson
             {
                 DataSources = app.GetDataSources()
                     .SelectMany(x => x.Value)
-                    .Where(x => !x.IsDataComponent)
+                    .Where(x => !x.IsDataComponent && x.TableDefinition == null)
```
lesaltzm, line 546: Instead of doing this in two passes, you could also put Int.MaxValue for the ones with table definitions in the OrderBy statement

src/PAModel/Serializers/MsAppSerializer.cs
MikeStall: Ie, something like:
        public int GetOrder(DataSourceEntry dataSource)
        {            
            return this.OrderDataSource.GetOrDefault<string, int>(dataSource.GetUniqueName(),
                (dataSource.TableDefinition == null) ? -1 : int.MaxValue
             );
        }

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -537,14 +537,22 @@ private static IEnumerable<FileEntry> GetMsAppFiles(this CanvasDocument app, Err
                 yield return ToFile(FileKind.PublishInfo, publishInfo);
 
             // "DataComponent" data sources are not part of DataSource.json, and instead in their own file
+            // In case the the entropy is missing then make sure that tableDefinitions are being added towards the end in DataSources.json file
```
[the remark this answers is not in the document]
MikeStall: [https://github.com/microsoft/PowerApps-Language-Tooling/issues/226](https://github.com/microsoft/PowerApps-Language-Tooling/issues/226)

MikeStall (approved):
:shipit:

src/PAModel/Serializers/MsAppSerializer.cs
MikeStall: test...

suskumar-MSFT: @MikeStall this change was reverted and a test case was added for this, do you want to add another test for something?

Issue #187, opened 2021-03-17
Working with the unicode screen names


vkusher:
The tool escapes all unicode chars and makes file's name unreadble.
The code also crashes when there is a "/" char in the screen's name
Consider to change the **DontEscapeChar** method inside of the **Utilities.cs**
to :
```
private static readonly List<char> invalidChars = Path.GetInvalidFileNameChars().ToList();

private static bool DontEscapeChar(char ch)
{
    return !invalidChars.Contains(ch);
}
```

MikeStall:
Are you on the latest builds? ("FormatVersion": "0.20", in the CanvasManifest.json). 

I just tried an app with a /, and it escapes it as  `Screen%2f1.fx.yaml`  
If it reproes on latest builds, can you share your msapp with us to repro?

One challenge for escaping is that we need to be cross-platform - the files may get unpacked on one OS, checked into git, and then used in another OS. So we can't use Path.GetInvalidFileNameChars.

vkusher:
You're right, the issue related to the lenght of the screen's name.

MikeStall:
How long is it? Can you share the screen's (unescaped) name here?

vkusher:
I think the problem appears when screen's name is greater then 40 unicode characters.
As an example, take this name:
одиндваодиндваодиндваодиндваодиндваодинд

MikeStall:
There are 2 issues here:
1. Long path names. Windows has a limit of MAX_PATH (260) characters for the full filename.  There may be some things to get around this. https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#win32-file-namespaces
2. Better handling of unicode. Do we escape to aggressively?  Escaping hits the first issue because escaping makes the path longer (a single high unicode char becomes 5 escape chars).

MikeStall:
We actually have more flexibility here for the yaml file name because the control's name is still inside the yaml contents. So the yaml file name is not needed to re-pack and get the control name. This means we could do something like truncate for long names, and add a mini hash to avoid collisions (use hash, not incrementing number)

MikeStall:
Adding Breaking tag since if we change the file encoding, that could be breaking.

PR #231, opened 2021-05-28
#187 Truncating the file names if length exceeds MAX_PATH

suskumar-MSFT:
It is more likely to happen when with unicode characters, if this happens - and the full path length exceeds 260 limit, the filname would be truncated and it would be appended with a hash at the end (to avoid duplicates) while retaining the file extension.

### DIFF
--- a/src/PAModel/Utility/DirectoryHelper.cs
+++ b/src/PAModel/Utility/DirectoryHelper.cs
@@ -69,7 +69,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
 
         public void WriteAllText(string subdir, FilePath filename, string text)
         {
-            string path = FilePath.ToValidPath(Path.Combine(_directory, subdir, filename.ToPlatformPath()));
+            string path = Path.Combine(_directory, subdir, filename.ToPlatformPath());
             EnsureFileDirExists(path);
             File.WriteAllText(path, text);
         }
--- a/src/PAModel/Utility/DirectoryHelper.cs
+++ b/src/PAModel/Utility/DirectoryHelper.cs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.AppMagic.Authoring.Persistence;
-using System;
 using System.IO;
 using System.Text.Json;
 using System.Linq;
--- a/src/PAModel/Utility/DirectoryHelper.cs
+++ b/src/PAModel/Utility/DirectoryHelper.cs
@@ -35,6 +35,12 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                     continue;
                 Directory.Delete(dir, recursive: true);
             }
+            foreach (var file in Directory.EnumerateFiles(_directory))
+            {
+                if (file.StartsWith(".git"))
+                    continue;
+                File.Delete(file);
+            }
         }
 
         public void WriteAllJson<T>(string subdir, FileKind kind, T obj)
@@ -70,6 +76,13 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         public void WriteAllText(string subdir, FilePath filename, string text)
         {
             string path = Path.Combine(_directory, subdir, filename.ToPlatformPath());
+
+            // Check for collision so that we don't overwrite an existing file.
+            if (File.Exists(path))
+            {
+                path = FilePath.ToFilePath(path).HandleFileNameCollisions(path);
+            }
+
             EnsureFileDirExists(path);
             File.WriteAllText(path, text);
         }
--- a/src/PAModel/Utility/DirectoryHelper.cs
+++ b/src/PAModel/Utility/DirectoryHelper.cs
@@ -71,7 +71,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
 
         public void WriteAllText(string subdir, FilePath filename, string text)
         {
-            string path = Path.Combine(_directory, subdir, filename.ToPlatformPath());
+            string path = FilePath.ToValidPath(Path.Combine(_directory, subdir, filename.ToPlatformPath()));
             EnsureFileDirExists(path);
             File.WriteAllText(path, text);
         }
--- a/src/PAModel/Utility/FilePath.cs
+++ b/src/PAModel/Utility/FilePath.cs
@@ -13,7 +13,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Utility
 {
     internal class FilePath
     {
-        public const int MAX_PATH = 260;
+        public const int MaxFileNameLength = 60;
         private const string yamlExtension = ".fx.yaml";
         private const string editorStateExtension = ".editorstate.json";
         private readonly string[] _pathSegments;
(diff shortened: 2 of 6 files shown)


src/PAModel/Utility/FilePath.cs
```diff
@@ -13,6 +13,9 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Utility
 {
     internal class FilePath
     {
+        private const int MAX_PATH = 260;
```
lesaltzm, line 16: This is the max path for windows in general, but what if we're unpacking to a directory that already has a path that's, say, 100 char long? I'd suggest that we consider having a shorter max path on our end so we're not pushing the absolute limit.

lesaltzm:
Can we add some test cases to UtilityTests.cs?

src/PAModel/Utility/FilePath.cs
suskumar-MSFT: this is to test the length of the absolute path

src/PAModel/Utility/DirectoryHelper.cs
```diff
@@ -71,7 +71,7 @@ public void WriteAllXML(string subdir, FilePath filename, string xmlText)
 
         public void WriteAllText(string subdir, FilePath filename, string text)
         {
-            string path = Path.Combine(_directory, subdir, filename.ToPlatformPath());
+            string path = FilePath.ToValidPath(Path.Combine(_directory, subdir, filename.ToPlatformPath()));
```
lesaltzm, line 74: Oh, I see. you're truncating the whole path, not just the part we're responsible for

lesaltzm: What if the base + _subdir part of the path exceeds MaxPath. Is that possible?

suskumar-MSFT: yea I was thinking that, but I don't think thats possible, if that happens we won't be able to unpack anyways

src/PAModel/Utility/FilePath.cs
```diff
@@ -30,6 +33,29 @@ public string ToPlatformPath()
...
+        /// Checks if the path length is more than MAX_PATH (260)
+        /// then it truncates the path, and adds a hash to the filename while retaining the file extension.
+        /// </summary>
+        /// <param name="path">Full path of the file.</param>
+        /// <returns></returns>
+        public static string ToValidPath(string path)
+        {
+            if (path.Length > MAX_PATH)
+            {
+                var extension = path.EndsWith(yamlExtension)
```
MikeStall, line 46: >path.EndsWith(yamlExtension) [](http://example.com/codeflow?start=32&length=28)

Can we use Path.GetExtension() ?

src/PAModel/Utility/FilePath.cs
```diff
@@ -30,6 +33,29 @@ public string ToPlatformPath()
             return Path.Combine(_pathSegments.Select(Utilities.EscapeFilename).ToArray());
         }
 
+        /// <summary>
+        /// Checks if the path length is more than MAX_PATH (260)
+        /// then it truncates the path, and adds a hash to the filename while retaining the file extension.
+        /// </summary>
+        /// <param name="path">Full path of the file.</param>
+        /// <returns></returns>
+        public static string ToValidPath(string path)
```
MikeStall, line 42: >public static string ToValidPath(string path) [](http://example.com/codeflow?start=8&length=45)

stepping back, Every other tool must be hitting this problem ... is there a common solution?

src/PAModel/Utility/FilePath.cs
```diff
@@ -30,6 +33,29 @@ public string ToPlatformPath()
...
+        {
+            if (path.Length > MAX_PATH)
+            {
+                var extension = path.EndsWith(yamlExtension)
+                    ? yamlExtension
+                    : path.EndsWith(editorStateExtension)
+                    ? editorStateExtension
+                    : Path.GetExtension(path);
+
+                var hash = string.Format("{0:X}", path.GetHashCode());
```
MikeStall, line 52: >path.GetHashCode() [](http://example.com/codeflow?start=50&length=18)

Danger! GetHashCode is not deterministic across process restarts. We need a deterministic hash function. since this isn't crypto, we can just grab a 3-liner and use that.

src/PAModel/Utility/FilePath.cs
MikeStall: There's some very good discussion on Stackoverflow around here, [https://stackoverflow.com/questions/23588944/better-to-check-if-length-exceeds-max-path-or-catch-pathtoolongexception?noredirect=1&lq=1](https://stackoverflow.com/questions/23588944/better-to-check-if-length-exceeds-max-path-or-catch-pathtoolongexception?noredirect=1&lq=1)

src/PAModel/Utility/FilePath.cs
suskumar-MSFT: the extension we use is .fx.yaml, the GetExtension() method only return .yaml, so one option would be assume that .yaml is always .fx.yaml but then if ever we have a file with just .yaml extension that might break our tool.

src/PAModel/Utility/FilePath.cs
```diff
@@ -112,5 +122,44 @@ public override int GetHashCode()
...
+        /// <summary>
+        /// If the file name length is longer than 60, it is truncated and appeded with a hash (to avoid collisions).
+        /// Checks the length of the escaped file name, since its possible that the length is under 60 before excaping but goes beyond 60 later.
+        /// We do module by 1000 of the hash to limit it to 3 characters.
+        /// </summary>
+        /// <returns></returns>
+        private string GetTruncatedFileNameIfTooLong(string fileName)
+        {
+            var newFileName = Utilities.EscapeFilename(fileName);
+            if (newFileName.Length > 60)
```
MikeStall, line 134: >60 [](http://example.com/codeflow?start=37&length=2)

use the constant?

src/PAModel/Utility/FilePath.cs
```diff
@@ -112,5 +122,44 @@ public override int GetHashCode()
...
+            {
+                var extension = newFileName.EndsWith(yamlExtension)
```
MikeStall, line 136: >EndsWith [](http://example.com/codeflow?start=44&length=8)

casing?

src/PAModel/Utility/FilePath.cs
```diff
@@ -112,5 +122,44 @@ public override int GetHashCode()
             return ToMsAppPath().GetHashCode();
         }
 
+        /// <summary>
+        /// If the file name length is longer than 60, it is truncated and appeded with a hash (to avoid collisions).
```
MikeStall, line 126: >appeded [](http://example.com/codeflow?start=75&length=7)

typo

MikeStall (approved):
:shipit:

lesaltzm (approved):

src/PAModelTests/FilePathTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.IO;
+
+namespace PAModelTests
+{
+    [TestClass]
+    public class FilePathTests
+    {
+        // If the length of the escaped file name is beyond 60, then we truncate the length to limit it to 60.
+        [DataTestMethod]
+        [DataRow("Long*Name*File*Path*Validation*Tests***.fx.yaml", FilePath.MaxFileNameLength)]
```
lesaltzm, line 16: This should also test the hash, we should compare against an expected actual name.

PR #233, opened 2021-06-02
Changes to throw unhandled exceptions to the caller

suskumar-MSFT:
Made Changes to throw unhandled exceptions to the caller when intended.

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -178,14 +178,13 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 }
                 return document;
             }
-            catch (Exception e)
+            catch (DocumentException e)
             {
-                // Internal error - something was thrown without adding to the error container.
-                errors.InternalError(e);
-
-                if (!(e is DocumentException))
+                if (!errors.HasErrors)
                 {
-                    throw;
+                    // Internal error - something was thrown without adding to the error container.
+                    // Add at least one error
+                    errors.InternalError(e);
                 }
                 return null;
             }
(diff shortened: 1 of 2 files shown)


src/PAModel/CanvasDocument.cs
```diff
@@ -96,7 +96,7 @@ public class CanvasDocument
         /// </summary>
         /// <param name="fullPathToMsApp">path to an .msapp file</param>
         /// <returns>A tuple of the document and errors and warnings. If there are errors, the document is null.  </returns>
-        public static (CanvasDocument, ErrorContainer) LoadFromMsapp(string fullPathToMsApp)
+        public static (CanvasDocument, ErrorContainer) LoadFromMsapp(string fullPathToMsApp, bool throwOnException = false)
```
lesaltzm, line 99: This seems odd, why is it an option for the caller? If the caller doesn't want the exceptions, it can always wrap this in a 
`try {} catch { /* do nothing */ }`

src/PAModel/CanvasDocument.cs
```diff
@@ -180,6 +180,10 @@ private static CanvasDocument Wrapper(Func<CanvasDocument> worker, ErrorContaine
             }
             catch (Exception e)
             {
+                if (throwOnException && !(e is DocumentException))
+                {
+                    throw;
+                }
```
lesaltzm, line 186: this could just catch DocumentExceptions and let the others pass by, if we aren't allowing the caller to control the behavior. 
```
try { ...
}
catch (DocumentException e)
{
// swallow, use existing behavior
}
```

MikeStall:
>         private static CanvasDocument Wrapper(Func<CanvasDocument> worker, ErrorContainer errors)

With the caller's TryOperation wrapping, do we need this at all?


---
Refers to: src/PAModel/CanvasDocument.cs:169 in 9a68d7f. [](commit_id = 9a68d7f6edfd0bd369c990dd177bdbc8a7ab068a, deletion_comment = False)

MikeStall (approved):
:shipit:

MikeStall:
>         private static CanvasDocument Wrapper(Func<CanvasDocument> worker, ErrorContainer errors)

discussed offline. Still useful to catch DocumentException here

---
In reply to: [853378314](https://github.com/microsoft/PowerApps-Language-Tooling/pull/233#issuecomment-853378314) [](http://example.com/codeflow?ancestors=853378314)

---
Refers to: src/PAModel/CanvasDocument.cs:169 in 9a68d7f. [](commit_id = 9a68d7f6edfd0bd369c990dd177bdbc8a7ab068a, deletion_comment = False)

lesaltzm (approved):
🚢

PR #235, opened 2021-06-04
#236 : Changes to file name truncation and checksum calculation

suskumar-MSFT:
* Limited the name truncation logic only to controls
* Found a bug in the check sum comparison, fixed the same


src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -950,5 +953,39 @@ private static void AddTestSuiteControl(CanvasDocument app, BlockNode controlIR)
...
+            }
+            return newControlName;
+        }
+
+        /// <summary>
+        /// djb2 algorithm to compute the hash of a string
+        /// </summary>
+        /// <param name="str"></param>
+        /// <returns></returns>
+        private static ulong GetHash(string str)
```
MikeStall, line 980: >GetHash [](http://example.com/codeflow?start=29&length=7)

nit: 
addd comment "This must be determinstic and stable since it's used in file names". 
I'd add a unit test to verify that - hash a literal and compare to expected value (like what we do with checksums).

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -950,5 +953,39 @@ private static void AddTestSuiteControl(CanvasDocument app, BlockNode controlIR)
...
+        /// We do modulo by 1000 of the hash to limit it to 3 characters.
+        /// </summary>
+        /// <returns></returns>
+        private static string TruncateControlNameIfTooLong(string controlName)
+        {
+            var newControlName = Utilities.EscapeFilename(controlName);
+            if (newControlName.Length > MaxControlNameLength)
+            {
+                // limit the hash to 3 characters by doing a module by 1000
+                var hash = (GetHash(newControlName) % 1000).ToString("x3");
```
MikeStall, line 969: >1000 [](http://example.com/codeflow?start=54&length=4)

nit: Since you're printing as hex, could this be 16^3 (instead of 10^3)

src/PAModel/Checksum/ChecksumMaker.cs
```diff
@@ -189,7 +189,8 @@ internal static string ChecksumToString(byte[] bytes)
 
             foreach (var kv in _files.OrderBy(x => x.Key))
             {
-                hash.AppendData(kv.Value);
+                var fileNameInBytes = Encoding.ASCII.GetBytes(kv.Key);
+                hash.AppendData(fileNameInBytes.Concat(kv.Value).ToArray());
```
MikeStall, line 193: Why this change?
This looks like a change to the checksum algorithm, which can be significant (server computes checksum too).

src/PAModelTests/FilePathTests.cs
```diff
@@ -1,27 +0,0 @@
...
-
-using Microsoft.PowerPlatform.Formulas.Tools;
-using Microsoft.PowerPlatform.Formulas.Tools.Utility;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-using System.IO;
-
-namespace PAModelTests
-{
-    [TestClass]
-    public class FilePathTests
```
MikeStall, line 12: >FilePathTests [](http://example.com/codeflow?start=17&length=13)

Did you intend to delete this (loss of coverage)?
Is there another test to add to fill the gap?

src/PAModel/Checksum/ChecksumMaker.cs
suskumar-MSFT: we were calculating whole checksum based on the content of the files, not account for the different filenames due to which the issue that my changes caused slipped through. I made this change to calculate the checksum for content of the file along with filename to ensure everything match.

src/PAModelTests/FilePathTests.cs
suskumar-MSFT: the changes to FilePath were reverted so removed this one, I will add the coverage where it is currently moved to.

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: we are doing the modulo by 1000 before converting to hex, so the max value would 999 in decimal.

src/PAModel/Checksum/ChecksumMaker.cs
MikeStall: I see - we're not including kv.Key in the hash. 

can we just add:
`hash.AppendData(kv.Key);`

and avoid the concat/toArray?

MikeStall:
>         public static string Version = "C6";

If we do change checksum, we should update this to 'C7'

We also have a bunch of tests that will fail (which validate checksum stability). 
This friction is intentional - checksum changes are a big deal. 


---
Refers to: src/PAModel/Checksum/ChecksumMaker.cs:25 in 644f80c. [](commit_id = 644f80c3b32521d8e5fbbcbb582bc69bc5e27b73, deletion_comment = False)

src/PAModel/Serializers/SourceSerializer.cs
MikeStall: 999.ToString("x3") // "3e7"
(16*16*16-1).ToString("x3") // "fff"

MikeStall: basically, if the goal is to condense the hash to 3 hex digits, I think you should replace '1000' with (16*16*16-1)

src/PAModel/Utility/FilePath.cs
```diff
@@ -25,19 +25,21 @@ public FilePath(params string[] segments)
 
         public string ToMsAppPath()
         {
-            return string.Join("\\", _pathSegments);
+            var path = string.Join("\\", _pathSegments);
+
+            // Some paths mistakenly start with DirectorySepChar in the msapp,
```
MikeStall, line 30: >mistakenly [](http://example.com/codeflow?start=26&length=10)

is this a server bug in generating the app?

suskumar-MSFT: possibly I have copied over this logic from DirectoryHelper.cs, where we remove this '_' when we load app from sources. @lesaltzm do you know if this was a server bug?

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -828,19 +831,19 @@ private static void RestoreViewName(DataSourceEntry ds, string dataSetName)
             }
 
             // Write out of all the other state for roundtripping 
-            string extraContent = (topParentname ?? controlName) + ".editorstate.json";
+            string extraContent = (topParentname ?? newControlName) + ".editorstate.json";
```
lesaltzm, line 834: should this always be `newControlName` for editorState?

suskumar-MSFT: we always dump the editorstate data per top control, for AppTest control since we are sharding into individual TestSuites the editorState.json would still be 1.

MikeStall:
Filed https://github.com/microsoft/PowerApps-Language-Tooling/issues/236 to track this.

src/PAModel/Checksum/ChecksumMaker.cs
```diff
@@ -21,8 +21,11 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
     /// </summary>
     public class ChecksumMaker
     {
+        // If the checksum is generated with C6 version then treat it as if it's missing.
+        public static string VersionC6 = "C6";
```
lesaltzm, line 25: Could we make this more generic? Parse the version and if its <= the current one, treat it as if it's missing?

suskumar-MSFT: do we want to loosen up that much? Earlier we were very strict, we required the exact version

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -227,7 +228,11 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
                 // This is debug only. The server checksum is out of date with the client checksum
                 // The main checksum validation that matters is the repack after unpack
 #if DEBUG
```
MikeStall, line 230: Now that we have an "is older" check, can we remove #if Debug and do this all the time?

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -227,7 +228,11 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
...
-                if (app._checksum.ServerStampedChecksum != null && app._checksum.ServerStampedChecksum != currentChecksum.wholeChecksum)
+                // In case the server stamped checksum on the msapp is older than the client, treat it like its missing checksum.
+                var isNullOrOlderChecksum = app._checksum.ServerStampedChecksum == null
```
MikeStall, line 232: >app._checksum.ServerStampedChecksum == null [](http://example.com/codeflow?start=44&length=43)

nit: may be cleaner w/ short circuiting.

(app._checksum.ServerStampedChecksum == null) |||
(ChecksumMaker.GetChecksumVersion(app._checksum.ServerStampedChecksum) < ChecksumMaker.Version)

src/PAModel/Utility/Utilities.cs
```diff
@@ -358,5 +360,41 @@ public static string GetResourceRelativePath(ContentKind contentType)
...
+        /// <param name="name">The name that needs to be truncated.</param>
+        /// <returns>Returns the truncated name if the escaped version of it is longer than 50 characters.</returns>
+        public static string TruncateNameIfTooLong(string name)
+        {
+            var escapedName = EscapeFilename(name);
+            if (escapedName.Length > MaxNameLength)
+            {
+                // limit the hash to 3 characters by doing a module by 4096 (16^3)
+                var hash = (GetHash(escapedName) % 4096).ToString("x3");
+                escapedName = escapedName.Substring(0, MaxNameLength - hash.Length) + hash;
```
MikeStall, line 378: >hash [](http://example.com/codeflow?start=86&length=4)

nit:
Can we inject an underscore, makes it easier to realize it's been truncated and hashed.  Like:

+ "_" + hash;

"FileNameTruncated_fff"

MikeStall (dismissed):
:shipit:

src/PAModel/Serializers/MsAppSerializer.cs
suskumar-MSFT: makes sense

Issue #194, opened 2021-04-08
Issues recombining combo box control's MultiValueDelimiter property


ccolomier:
When merging two apps we ran into an issue where the MultiValueDelimiter property was returned as just a comma and space without the quotes. 

![image](https://user-images.githubusercontent.com/32962556/114096241-fdd08f00-98ad-11eb-8947-e2a08ce386b9.png)

The value should be, based on our original app. 

![image](https://user-images.githubusercontent.com/32962556/114096330-1b055d80-98ae-11eb-83c0-a3178ad3ff7c.png)

We can't find a reason as to why it was changed from ", " to just a comma and space character.

lesaltzm:
Can you post/screenshot the .fx.yaml files after you merged the apps?

ccolomier:
Good call,

It doesn't seem to even have the MultiValueDelimiter property in the .fx.yaml file:

![image](https://user-images.githubusercontent.com/32962556/114099611-9d901c00-98b2-11eb-8fbe-01a0b6981b11.png)

lesaltzm:
Is ToolTypeInput.MultiValueDelimiter present in either of the apps you were merging?

ccolomier:
I don't believe it is. The original app, unpacked using this tool again, produced the following .fx.yaml code which defines the ToolTypeInput comobox control:

![image](https://user-images.githubusercontent.com/32962556/114188365-1032e300-9917-11eb-93d4-2b6b3c0f4196.png)


Obviously, as you can see, ToolTypeInput.MultiValueDelimiter is also missing from the original unpacked app.

PR #239, opened 2021-06-07
#194 : Changes to handle string type properties.

suskumar-MSFT:
While parsing control property fs the property is of string type and 
* not an expression
* not a localizedValue
* Not a uri

Then escape double quotes to retain string literal values


lesaltzm (approved):
🚢

PR #242, opened 2021-06-08
Fix #238 - better handling of user errors.

### LINKED ISSUE #238: Better handling of user errors on input arguments
The key requirement is that: PAC CLI doesn't report crashes for user errors, only report crashes for our actual bugs.

The common user errors here would be bad arguments, such as file paths to missing files. Or a corrupt zip. More esoteric ones would be a locked file

PAC CLI reports crashes for any unhandled exception. Today, these bad args cause unhandled exceptions.

Things we report back through our ErrorCodes are considered user error and not our bug.

We have a few ways we can fix:

1. The CanvaDocument.Load|Save methods can catch these common errors and return a new ErrorCode. (Toiday, they just throw exceptions). 
2. We continue to throw, but CLI catches before going unhandled. --> this can be a lot of catches (InvalidOp, Arg Exeption, IO exception, etc). 
3. CLI validates arguments before calling CanvasDocument. (and hence avoiding the exceptions). --> not ideal.

Recommending 1.

MikeStall:
Fix #238

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -23,6 +23,11 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
     /// </summary>
     public class CanvasDocument
     {
+        /// <summary>
+        /// Current source format version. 
+        /// </summary>
+        public Version CurrentSourceVersion => SourceSerializer.CurrentSourceVersion;
+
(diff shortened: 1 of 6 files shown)


lesaltzm (approved):
🚢

suskumar-MSFT (approved):

PR #245, opened 2021-06-10
Chnages to handle InvarantScript for function parameters

suskumar-MSFT:
In certain cases the value of InvariantScript of the function parameter's rule entry is not same as its default value in the template. We were always using the DefaultValue from the template to generate the InvariantScript's value. Added handling to store this information explicitly if present.


src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -294,7 +301,10 @@ private static (ControlInfoJson.Item item, int index) CombineIRAndState(BlockNod
                             if (arg.Identifier == PAConstants.ThisPropertyIdentifier)
                                 continue;
 
-                            properties.Add(GetPropertyEntry(state, errors, funcName + "_" + arg.Identifier, arg.Default.Expression));
+                            // Use InvariantScript expression if present otherwise use Default expression.
+                            var propName = funcName + "_" + arg.Identifier;
+                            var script = entropy.FunctionParamsInvariantScripts.ContainsKey(propName) ? entropy.FunctionParamsInvariantScripts[propName] : arg.Default.Expression;
```
MikeStall, line 306: >entropy.FunctionParamsInvariantScripts [](http://example.com/codeflow?start=41&length=38)

nit: can we use GetOrDefault()  ? 


entropy.FunctionParamsInvariantScripts.GetOrDefault(propName, arg.Default.Expression);

src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -89,6 +89,13 @@ private static void SplitIRAndState(ControlInfoJson.Item control, string topPare
                                 },
                             });
 
+                            // Handle the case where invariantScript value of the property is not same as the default script.
+                            var invariantScript = control.Rules.First(rule => rule.Property == arg.Name)?.InvariantScript;
+                            if (invariantScript != null && invariantScript != arg.ScopeVariableInfo.DefaultRule)
```
MikeStall, line 94: > if (invariantScript != null && invariantScript != arg.ScopeVariableInfo.DefaultRule) [](http://example.com/codeflow?start=27&length=85)

Is this too broad? It seems like it will catch all the expressions and put everything into entropy?

src/PAModel/Entropy.cs
```diff
@@ -238,5 +241,23 @@ public bool TryGetDataTableControlJson(string controlName, out string json)
 
             return DataTableCustomControlTemplateJsons.TryGetValue(controlName, out json);
         }
+
+        public string GetDefaultScript(string propName, string defaultValue)
+        {
+            if(FunctionParamsInvariantScripts.TryGetValue(propName, out string[] value) && value?.Length == 2)
```
lesaltzm, line 247: nit: `if (`

src/PAModel/Entropy.cs
```diff
@@ -58,6 +58,9 @@ internal class PropertyEntropy
         // Key is resource name, value is filename
         public Dictionary<string, string> LocalResourceFileNames { get; set; } = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
+        // Some of the properties have different value of InvariantScript than the DefaultScript which should be the case, hence need special handling.
```
lesaltzm, line 61: maybe "Some Component Function Parameter Properties are serialized with a different InvariantScript and DefaultScript. We show the InvariantScript in the .yaml, and persist the difference in Entropy to ensure we have accurate roundtripping behavior"

src/PAModel/Entropy.cs
```diff
@@ -58,6 +58,9 @@ internal class PropertyEntropy
...
+        public Dictionary<string, string[]> FunctionParamsInvariantScripts { get; set; } = new Dictionary<string, string[]>(StringComparer.OrdinalIgnoreCase);
```
lesaltzm, line 62: is string[] correct? Couldn't we just persist a dict of <propName, defaultScript>, since the invariant script is in the Yaml?

src/PAModel/Entropy.cs
lesaltzm: Also, call out what the key and value are in the comment.

src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -294,7 +301,9 @@ private static (ControlInfoJson.Item item, int index) CombineIRAndState(BlockNod
                             if (arg.Identifier == PAConstants.ThisPropertyIdentifier)
                                 continue;
 
-                            properties.Add(GetPropertyEntry(state, errors, funcName + "_" + arg.Identifier, arg.Default.Expression));
+                            var propName = funcName + "_" + arg.Identifier;
+                            properties.Add(GetPropertyEntry(state, errors, propName, entropy.GetUserScript(propName, arg.Default.Expression)));
+                            arg.Default.Expression = entropy.GetDefaultScript(propName, arg.Default.Expression);
```
lesaltzm, line 306: Why are we modifying the IR here? If it's to pass the default script value into RepopulateTemplateCustomProperties, that seems a bit fragile. There's probably a better way to do that.

src/PAModel/Entropy.cs
lesaltzm: Discussed offline, Sushil to add comment explaining why this is the case.

lesaltzm (approved):
Approved. If this is testable I'd love to see tests, otherwise, make sure we have representative apps in our internal suite.

Issue #244, opened 2021-06-10
Validate before deleting directories


MikeStall:
When unpacking, we delete the target directory - this is so stale artifacts aren't overwritten. 
But if they specify the wrong directory (ie, desktop), that could delete too much and cause major data loss.

MikeStall:
See call to DeleteAllSubdirs()
https://github.com/microsoft/PowerApps-Language-Tooling/blob/58275118c9aa429e6605bf734b95926be81c4196/src/PAModel/Serializers/SourceSerializer.cs#L443

PR #246, opened 2021-06-11
#244 : Validate before we delete the directory on unpack

suskumar-MSFT:


src/PAModel/Utility/DirectoryHelper.cs
```diff
@@ -29,17 +29,20 @@ public void DeleteAllSubdirs()
             {
                 Directory.CreateDirectory(_directory);
             }
-            foreach (var dir in Directory.EnumerateDirectories(_directory))
+            if (IsValidForDelete())
```
lesaltzm, line 32: Should we report an error and throw if the directory isn't valid for deletion?

lesaltzm (approved):

src/PAModel/Utility/DirectoryHelper.cs
MikeStall: Yes, I think we should.  Because after unpack, they're going to repack, and we don't want to include a bunch of unrecognized files. 

The scenario here is that if they run it on the same git enabled folder each time, it should work ok. 

But if they run it on desktop, it should error.

src/PAModel/Utility/DirectoryHelper.cs
```diff
@@ -129,6 +132,20 @@ public bool FileExists(string subdir, string filename)
             string path = Path.Combine(_directory, subdir, filename);
             return File.Exists(path);
         }
+
+        /// <summary>
+        /// Returns true if it's either an empty directory or it contains CanvasManifest.json file.
+        /// </summary>
+        /// <returns></returns>
+        private bool IsValidForDelete(ErrorContainer errors)
```
MikeStall, line 140: >bool IsValidForDelete [](http://example.com/codeflow?start=16&length=21)

nit: since this either returns true or Throws (never returns false), it could be:
void ValidateSafeToDelete(errors)

MikeStall (approved):
:shipit:

PR #247, opened 2021-06-16
AppTest control missing properties when converted to source

suskumar-MSFT:
Added handling to shard the test suites into individual files keep the root control file in place as well to write the additional properties.


src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -375,9 +375,18 @@ private static void LoadSourceFiles(CanvasDocument app, DirectoryReader director
...
+            // Make sure to load the the Test_7F478737223C4B69.fx.yaml file first to add the top parent control.
+            var appTestControl = directory.EnumerateFiles(TestDir, AppTestControlName + ".fx.yaml").FirstOrDefault();
+            if (appTestControl != null)
+            {
+                AddControl(app, appTestControl._relativeName, false, appTestControl.GetContents(), errors);
+            }
             foreach (var file in directory.EnumerateFiles(TestDir, "*.fx.yaml"))
             {
-                AddControl(app, file._relativeName, false, file.GetContents(), errors);
+                if (file._relativeName != AppTestControlName + ".fx.yaml")
```
MikeStall, line 386: possible casing bug (if filenames are case insensitive)?

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -375,9 +375,18 @@ private static void LoadSourceFiles(CanvasDocument app, DirectoryReader director
             }
 
             // When loading TestSuites sharded files, add them within the top parent AppTest control (i.e. Test_7F478737223C4B69)
+            // Make sure to load the the Test_7F478737223C4B69.fx.yaml file first to add the top parent control.
```
MikeStall, line 378: Is there an assumption about control ordering here? Why does the order of AddControl matter?

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: This the top parent control for AppTest which has fixed name "Test_7F478737223C4B69"

src/PAModel/Serializers/SourceSerializer.cs
suskumar-MSFT: Earlier we had dissembled AppTest control and sharded each of the TestSuite inside it into individual files. Yesterday while investigating one of the telemetry bugs, I got hold of an app which had some properties defined at the root level, and since we were not persisting the top parent those properties were getting lost.

With this change we are still sharding TestStuies into individual files but we are still keeping the top parent (Test_7F478737223C4B69) file as well to persist root level properties if any.

When adding control we have to add the top parent control first so that we can append the children to it, hence this handling.

lesaltzm: Could we add a case that demonstrates this to sourcebaseline tests?

suskumar-MSFT: sure will do.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -375,9 +375,18 @@ private static void LoadSourceFiles(CanvasDocument app, DirectoryReader director
             }
 
             // When loading TestSuites sharded files, add them within the top parent AppTest control (i.e. Test_7F478737223C4B69)
```
lesaltzm, line 377: I'm wondering whether this code belongs here, or in the CanvasDocument.ApplyBefore/AfterLoad functions

lesaltzm: Any thoughts?

suskumar-MSFT: I am not sure if this can be moved there, since its happening at the time of reading the source files. We read the source files and we add controls to add it to the app object. Once we have the app object then we perform ApplyBefore/AfterLoad functions in the SaveAsMsApp method.

lesaltzm: Fair enough.

lesaltzm (approved):

src/PAModel/Serializers/SourceSerializer.cs
MikeStall: My concern is that the file system is case insensitive, and so could it somehow loses case  (becomes "TEST_7F..."). Basically, should this be a string insensitive compare? Or should we have a higher level helper for comparing filenames? (our "FileKind" sort of handles that)

suskumar-MSFT: updated the code to use FileKind

MikeStall (approved):
:shipit:

PR #253, opened 2021-06-25
Fix 252

### LINKED ISSUE #252: Unpack failing with PA3001: Internal error. Sequence contains no matching element
We have a large .msapp PowerApp that is failing to unpack using the PowerPlatform CLI v1.7.2 version of the tool.

Is that version of the CLI using the latest release from this repository (i.e. "0.3.1-preview")?

The error message we receieve is very generic:
```
Sorry, the app encountered a non recoverable error and will need to terminate. The exception details have been captured and will be forwarded to the development team, if telemetry has been enabled. Exception Id:b7429ee6-1ea7-44fd-b57f-96e9ca44ba83
```

Is there any functionality/command switch we can use to gain a more detailed verbose error message?

MikeStall:
Fix #252 
Controls don't necessarily have rules for all CustomProperties in a control.

Note after this, we still hit another issue with InvariantScript not round tripping (possibly #254)

### DIFF
--- a/src/PAModel/IR/IRStateHelpers.cs
+++ b/src/PAModel/IR/IRStateHelpers.cs
@@ -55,7 +55,13 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                     {
                         var name = customProp.Name;
                         customPropsToHide.Add(name);
-                        var expression = control.Rules.First(rule => rule.Property == name).InvariantScript;
+                        var rule = control.Rules.FirstOrDefault(rule => rule.Property == name);
+                        if (rule == null)
+                        {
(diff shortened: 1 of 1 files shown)


suskumar-MSFT (approved):

MikeStall:
The crash was:

                    // Create FunctionNodes on def
                    foreach (var customProp in control.Template.CustomProperties.Where(prop => prop.IsFunctionProperty))
                    {
                        var name = customProp.Name;
                        customPropsToHide.Add(name);
                        var expression = control.Rules.First(rule => rule.Property == name).InvariantScript;

PR #255, opened 2021-06-25
Bug254

### LINKED ISSUE #254: yaml round trip error
Getting this error from an msapp:
`Error   PA3003: Parse error: Property should be at same indent level at Workflows.fx.yaml:2285,17-2285,17`

MikeStall:
Fix #254 
Round trip \r\n pattern.

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlWriter.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlWriter.cs
@@ -110,6 +110,13 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
                         WriteIndent();
                         needIndent = false;
                     }
+
+                    if (ch == '\r')
+                    {
+                        // skip and handle at the \n
+                        continue; 
+                    }
+
                     if (ch == '\n')
                     {
                         _text.WriteLine();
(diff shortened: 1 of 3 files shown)


src/PAModel/PAConvert/Yaml/YamlWriter.cs
```diff
@@ -110,6 +110,13 @@ public void WriteProperty(string propertyName, string value)
                         WriteIndent();
                         needIndent = false;
                     }
+
+                    if (ch == '\r')
```
lesaltzm, line 114: \r shouldn't be making it to the yamlwriter, see a couple places in IRStateHelper where we do something like  
`           var script = rule.InvariantScript.Replace("\r\n", "\n").Replace("\r", "\n").TrimStart();`

What kind of property is hitting this and why is `\r\n` making its way into the IR to be written?

lesaltzm: Maybe the right thing to do here is to define an actual constructor for ExpressionNode and do this trimming/normalizing there.

MikeStall: The yaml layer should handle \r  (we can't rely on the upstream caller to remove them).  Ie, see new test added in this pr. 
In this bug, there's probably some path that's skipping the IR's replacement.

MikeStall: We're definitely missing a bunch of spots in the IRStateHelper. 
Although if we do it in the lexer, then we should be able to completely remove the conversions from IRStateHelper . I'll try that ...

lesaltzm (approved):

PR #263, opened 2021-07-06
Make version property static for easier discoverability by consumers.

MikeStall:
It was an instance property on CanvasDocument, but that class doesn't have a public ctor either.


lesaltzm (approved):
🚢

suskumar-MSFT (approved):

PR #265, opened 2021-07-06
unrecognized hex char fix

suskumar-MSFT:
Each EscapeChar (%) must be followed by two integer values if that is not the case then it is likely that the truncation left incomplete escapes. So making a fix to remove if the string ends with % or % followed by a single digit.

### DIFF
--- a/src/PAModel/Utility/Utilities.cs
+++ b/src/PAModel/Utility/Utilities.cs
@@ -383,6 +383,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
 
         /// <summary>
         /// Truncates a string with the given length and strips off incomplete escape characters if any.
+        /// Each EscapeChar (%) must be followed by two integer values if that is not the case then it is likely that the truncation left incomplete escapes.
         /// </summary>
         /// <param name="name">The string to be truncated</param>
         /// <param name="length">The max length of the truncated string.</param>
--- a/src/PAModel/Utility/Utilities.cs
+++ b/src/PAModel/Utility/Utilities.cs
@@ -376,7 +376,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 // limit the hash to 3 characters by doing a module by 4096 (16^3)
                 var hash = (GetHash(escapedName) % 4096).ToString("x3");
                 escapedName = TruncateName(escapedName, MaxNameLength - hash.Length -1) + "_" + hash;
-
             }
             return escapedName;
         }
@@ -390,7 +389,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         /// <returns>Truncated string.</returns>
         private static string TruncateName(string name, int length)
         {
-            int removeTrailingCharsLength = name[length - 1] == EscapeChar ? 1 : name[length - 2] == EscapeChar ? 2 : 0;
+            int removeTrailingCharsLength = name[length - 1] == EscapeChar ? 1 : (name[length - 2] == EscapeChar ? 2 : 0);
             return name.Substring(0, length - removeTrailingCharsLength);
         }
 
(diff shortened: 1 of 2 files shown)


src/PAModel/Utility/Utilities.cs
```diff
@@ -375,11 +375,25 @@ public static string TruncateNameIfTooLong(string name)
...
         }
 
+        /// <summary>
+        /// Truncates a string with the given length and strips off incomplete escape characters if any.
+        /// Each EscapeChar (%) must be followed by two integer values if that is not the case then it is likely that the truncation left incomplete escapes.
+        /// </summary>
+        /// <param name="name">The string to be truncated</param>
+        /// <param name="length">The max length of the truncated string.</param>
+        /// <returns>Truncated string.</returns>
+        private static string TruncateName(string name, int length)
```
MikeStall, line 391: >TruncateName [](http://example.com/codeflow?start=30&length=12)

Is there some unit test to add?

src/PAModel/Utility/Utilities.cs
```diff
@@ -375,11 +375,25 @@ public static string TruncateNameIfTooLong(string name)
...
+        {
+            int removeTrailingCharsLength = name[length - 1] == EscapeChar ? 1 : name[length - 2] == EscapeChar ? 2 : 0;
```
lesaltzm, line 393: Very minor nit, but adding parens would make this nested ternary much more readable.

src/PAModel/Utility/Utilities.cs
```diff
@@ -375,11 +375,25 @@ public static string TruncateNameIfTooLong(string name)
             {
                 // limit the hash to 3 characters by doing a module by 4096 (16^3)
                 var hash = (GetHash(escapedName) % 4096).ToString("x3");
-                escapedName = escapedName.Substring(0, MaxNameLength - (hash.Length + 1)) + "_" + hash;
+                escapedName = TruncateName(escapedName, MaxNameLength - hash.Length -1) + "_" + hash;
+
```
lesaltzm, line 379: nit: extra newline

src/PAModel/Utility/Utilities.cs
MikeStall, line 393: >name[length - 1] == EscapeChar ? 1 : name[length - 2] == EscapeChar ? 2 : 0; [](http://example.com/codeflow?start=44&length=76)

nit: can you put parenthesis in here to disambiguate?

src/PAModel/Utility/Utilities.cs
suskumar-MSFT: There was an existing unit test which I have updated as needed

src/PAModel/Utility/Utilities.cs
```diff
@@ -375,11 +375,24 @@ public static string TruncateNameIfTooLong(string name)
...
+        /// <summary>
+        /// Truncates a string with the given length and strips off incomplete escape characters if any.
+        /// Each EscapeChar (%) must be followed by two integer values if that is not the case then it is likely that the truncation left incomplete escapes.
+        /// </summary>
+        /// <param name="name">The string to be truncated</param>
+        /// <param name="length">The max length of the truncated string.</param>
+        /// <returns>Truncated string.</returns>
+        private static string TruncateName(string name, int length)
+        {
+            int removeTrailingCharsLength = name[length - 1] == EscapeChar ? 1 : (name[length - 2] == EscapeChar ? 2 : 0);
```
MikeStall, line 392: >name[length - 1]  [](http://example.com/codeflow?start=44&length=17)

what if name is just 1 char?

suskumar-MSFT: currently this only gets invoked if the filename is longer than 50 chars

MikeStall: resolved offline: check on line 374 protects us.

MikeStall (approved):
:shipit:

PR #266, opened 2021-07-07
screens with excaped chars fail if test steps have references for it

suskumar-MSFT:
Fix for below error and similar types of errors

Error   PA3005: Validation error: Test Step Step22 references screen CheckBox,Toggle,Slider and Rating that is not present in the app


lesaltzm (approved):
🚢

PR #270, opened 2021-07-09
Handling for components in LDR

suskumar-MSFT:


MikeStall:
>                 }

This code is a problem - we're taking LDR and matching it to datasources, and then only serializing the datasources (and not LDR). 

We hit issues when an LDR entry doesn't match to Datasources (it gets ignored in serialization)

---
Refers to: src/PAModel/Serializers/SourceSerializer.cs:691 in 43ee46e. [](commit_id = 43ee46e5e37f19fc26e3a983045b8dc3b124f78b, deletion_comment = False)

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -694,6 +695,20 @@ private static void WriteDataSources(DirectoryWriter dir, CanvasDocument app, Er
 
                 dir.WriteAllJson(DataSourcesDir, new FilePath(filename), dataSourceStateToWrite);
             }
+
+            if (app._dataSourceReferences != null)
+            {
+                foreach (var kvp in app._dataSourceReferences)
+                {
+                    if (kvp.Value?.components?.Any() == true)
```
MikeStall, line 703: >if (kvp.Value?.components?.Any() == true) [](http://example.com/codeflow?start=20&length=41)

This may be too specific - 
can we just make this a catch-all (ie, regardless of whether it's a component, anything not captured by above L682-L691 gets written here)

MikeStall (approved):
:shipit:

PR #271, opened 2021-07-13
Properly merge added non-default properties

lesaltzm:


src/PAModel/MergeTool/Deltas/ChangeProperty.cs
```diff
@@ -73,6 +73,9 @@ public override void Visit(BlockNode node, ControlPath context)
                             return;
                         }
                     }
+
+                    // Property wasn't present in base
```
suskumar-MSFT, line 77: > // Property wasn't present in base [](http://example.com/codeflow?start=19&length=35)

when does this happen ?

suskumar-MSFT (approved):
:shipit:

MikeStall (approved):
:shipit:

src/PAModel/MergeTool/Deltas/ChangeProperty.cs
lesaltzm: This is if it was a default-valued property in the base, since we remove those from the IR.

Issue #249, opened 2021-06-18
Deleting all .editorstate files when the app has Tests leads to a validation error on -pack


lesaltzm:
We use the EditorState to convert screen names to the corresponding UniqueId: 
![image](https://user-images.githubusercontent.com/69215460/122591974-8d19c180-d018-11eb-91c4-c6bbd3548450.png)
We should create a path for this to work when statestore.contents is empty.
Otherwise, pack hits the error:                        
` _errors.ValidationError($"ScreenId referenced by TestStep {testStep.Rule} in {control.Name.Identifier} could not be found");`

PR #272, opened 2021-07-13
#201 & #249 : Screen references in tests break if entropy or editorstates are deleted

suskumar-MSFT:
Made the changes to read the screen names from app._screens rather than reading them from stateStore.
Also updated the logic to store screen index when entropy is deleted

### DIFF
--- a/src/PAModel/SourceTransforms/AppTestTransform.cs
+++ b/src/PAModel/SourceTransforms/AppTestTransform.cs
@@ -34,7 +34,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.SourceTransforms
             return templateName == "TestSuite";
         }
 
-        public AppTestTransform(ErrorContainer errors, TemplateStore templateStore, EditorStateStore stateStore, Entropy entropy)
+        public AppTestTransform(CanvasDocument app, ErrorContainer errors, TemplateStore templateStore, EditorStateStore stateStore, Entropy entropy)
         {
             _testStepTemplateName = "TestStep";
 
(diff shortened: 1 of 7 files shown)


src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -43,9 +43,8 @@ public AppTestTransform(ErrorContainer errors, TemplateStore templateStore, Edit
                 _testStepTemplateName = "TestStep" + i;
 
             var idRestorer = new UniqueIdRestorer(entropy);
-            _screenIdToScreenName = stateStore.Contents
-                .Where(state => state.TopParentName == state.Name)
-                .Select(state => new KeyValuePair<string, string>(idRestorer.GetControlId(state.Name).ToString(), state.Name)).ToList();
+            _screenIdToScreenName = app._screens
+                .Select(screen => new KeyValuePair<string, string>(idRestorer.GetControlId(screen.Key).ToString(), screen.Key)).ToList();
```
lesaltzm, line 47: Should the idRestorer be a property on CanvasDocument, if it's tracking global mapping? instead of instantiating it here.

suskumar-MSFT: We probably can, we use idRestorer at two places currently, one here and the other in IRStateHelper...The IRStateHelper currently doesn't take in CanvasApp object so idRestorer won't be available there, are you suggesting we start passing CanvasApp object to IRStateHelper as well?

suskumar-MSFT: ignore my comment, I just saw that we are already passing the idRestorer to IRStateHelper, so definitely makes sense for this to be added to CanvasDocument, I ll make the change

lesaltzm:
Could we test this somehow?

suskumar-MSFT:
Added tests

---
In reply to: [879359952](https://github.com/microsoft/PowerApps-Language-Tooling/pull/272#issuecomment-879359952) [](http://example.com/codeflow?ancestors=879359952,879359952)

lesaltzm (approved):

PR #274, opened 2021-07-14
Add == comparison for control paths

lesaltzm:


src/PAModel/Utility/ControlPath.cs
```diff
@@ -43,6 +43,10 @@ public ControlPath(List<string> segments)
             _segments = segments;
         }
 
+        public static bool operator ==(ControlPath left, ControlPath right) => left?.Equals(right) ?? right == null;
```
MikeStall, line 46: >== [](http://example.com/codeflow?start=108&length=2)

is this recursive?
Maybe cast to (object) or Object.ReferenceEquals

src/PAModel/Utility/ControlPath.cs
```diff
@@ -43,6 +43,10 @@ public ControlPath(List<string> segments)
...
+
+        public static bool operator !=(ControlPath left, ControlPath right) => !(left?.Equals(right) ?? right == null);
```
MikeStall, line 48: >!(left?.Equals(right) ?? right == null); [](http://example.com/codeflow?start=79&length=40)

Can this just be: 
!(left==right)

MikeStall (approved):
:shipit:

PR #278, opened 2021-07-15
Null handling when there is no data source in LocalConnectionReferences

suskumar-MSFT:


src/PAModel/CanvasDocument.cs
```diff
@@ -492,13 +492,16 @@ internal void OnLoadComplete(ErrorContainer errors)
                     errors.FormatNotSupported($"Document consistency error. Connection id mismatch");
                     throw new DocumentException();
                 }
-                foreach (var dataSourceName in connection.dataSources)
+                if (connection.dataSources != null)
                 {
-                    var ds = _dataSources.SelectMany(x => x.Value).Where(x => x.Name == dataSourceName).FirstOrDefault();
-                    if (ds == null)
+                    foreach (var dataSourceName in connection.dataSources)
```
lesaltzm, line 497: I generally find writing this as 
`foreach (var dataSourceName in connection.dataSources ?? Enumerable.Empty())` 
to be a bit cleaner, but it depends on if you want to emphasize the null check or not.

lesaltzm (approved):

src/PAModel/CanvasDocument.cs
MikeStall: We have an extension method (NullOk<T> ) for this too, you could writer...  

foreach (var dataSourceName in connection.dataSources.NullOk())

(see line 486). Turns out extension methods can accept null :D

MikeStall (approved):
:shipit:

MikeStall (approved):
:shipit:

PR #287, opened 2021-07-27
Add Theme and Setting Smart Merge support

lesaltzm:


emkessle (approved):
Consider keeping the if { } styles consistent. Sometimes you use brackets and sometimes you  don't

suskumar-MSFT (approved):

PR #295, opened 2021-08-01
add some stress tests

MikeStall:
getting failures, but using PR to comment.

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -309,6 +309,8 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             _logoFile = other._logoFile != null ? new FileEntry(other._logoFile) : null;
             _entropy = other._entropy.JsonClone();
             _checksum = other._checksum.JsonClone();
+
+            this._idRestorer = new UniqueIdRestorer(this._entropy);
         }
 
         // iOrder is used to preserve ordering value for round-tripping. 
--- a/src/PAModel/Checksum/ChecksumMaker.cs
+++ b/src/PAModel/Checksum/ChecksumMaker.cs
@@ -299,6 +299,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                             }
                             else
                             {
+                                hash.AppendNull();
                             }
                         }
                     }
--- a/src/PAModel/Checksum/ChecksumMaker.cs
+++ b/src/PAModel/Checksum/ChecksumMaker.cs
@@ -299,7 +299,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                             }
                             else
                             {
-                                hash.AppendNull();
+
                             }
                         }
                     }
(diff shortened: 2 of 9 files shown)


src/PAModel/CanvasDocument.cs
```diff
@@ -309,6 +309,8 @@ internal CanvasDocument(CanvasDocument other)
             _logoFile = other._logoFile != null ? new FileEntry(other._logoFile) : null;
             _entropy = other._entropy.JsonClone();
             _checksum = other._checksum.JsonClone();
+
+            this._idRestorer = new UniqueIdRestorer(this._entropy);
```
MikeStall, line 313: This looks like a real bug. 
The regular load paths allocate this, was missing from ctor path, and causing null refs.

src/PAModel/MsAppTest.cs
```diff
@@ -14,6 +16,126 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
...
+            using (var temp2 = new TempFile())
+            {
+                doc1.SaveToMsApp(temp1.FullPath);
+                doc2.SaveToMsApp(temp2.FullPath);
+
+                var doc1NoEntropy = RemoveEntropy(temp1.FullPath);
+                var doc2NoEntropy = RemoveEntropy(temp2.FullPath);
+
+                Compare(doc1NoEntropy, doc2NoEntropy, Console.Out);
+            }
```
MikeStall, line 82: Getting tons of errors from this test.

src/PAModel/MsAppTest.cs
```diff
@@ -14,6 +16,126 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
...
+                errors.ThrowOnErrors();
+
+                
+#if false
+                // $$$ We get different checksums when deleting in-memory vs. files.
+                // In some cases, files are just rearranged (\images),
+                // but in other cases, look different. 
+
+                // Reset Entropy on in-memory model
+                doc1._entropy = new Entropy();
```
MikeStall, line 115: This constantly fails too. The expectation (perhaps wrongly?) is that clearing the entropy in memory would have same affect as deleting entropy on disk. 
In reality, lots of unexpected diffs.

src/PAModel/CanvasDocument.cs
MikeStall: actually, I think this still has an issue, in Merger.ApplyDelta(), we delete entropy via setting to null. That means idRestorer would be pointing to an orphaned entropy object.

src/PAModel/Checksum/ChecksumMaker.cs
```diff
@@ -299,6 +299,7 @@ private static void ChecksumJson(Context ctx, IHashMaker hash, JsonElement je)
                             }
                             else
                             {
+                                hash.AppendNull();
```
MikeStall, line 302: >hash.AppendNull(); [](http://example.com/codeflow?start=32&length=18)

Will this require a new hash C9?

lesaltzm: Yeah, but actually I think I can fix it without doing that. The bug is with DebugTextHashMaker, if you write an object like 
`{key: null, otherKey: 123}`, the json writer complains that you wrote two keys in a row.

MikeStall:
Will submit more recent version.

PR #297, opened 2021-08-02
fix bug 282 - roundtrip was failing.

MikeStall:


MikeStall:
The difference is what happens when FirstParty flag is missing (Which happens on older apps).

"Bool" means that we always write "FirstParty : false".
"Bool?" means that we omit it altogether - which is consistent with what older apps would do before the flag exists.

lesaltzm (approved):
:ship:

PR #302, opened 2021-08-05
Pick a winner for screen order changes

lesaltzm:


MikeStall:
Is ScreenOrder special - do we hit this problem with any ordered collection?

lesaltzm:
> Is ScreenOrder special - do we hit this problem with any ordered collection?

I haven't seen this with any other ordered collections (most of which are in /Entropy). Screen order is one of the few with some impact on the resulting app, so it's important to merge correctly.

MikeStall (approved):

PR #316, opened 2021-11-02
Allow Visual Studio project and solution files to be present in the extraction directory

### LINKED ISSUE #314: Requirement of an empty directory for unpacking .msapp file
It appears that there is a requirement when unpacking an .msapp file that the destination folder must be empty of all files, or contain a 'CanvasManifest.json' file (see [here](https://github.com/microsoft/PowerApps-Language-Tooling/blob/d8ee30c12de93eebe784b5de1b4d4ae3528589c0/src/PAModel/Utility/DirectoryHelper.cs#L140-L148)). Is there any way to not make this a requirement?  I would like to extract the contents of the app at the root folder of an existing Visual Studio project which will always contain at the very least the '.csproj' file.  In order to work around this limitation I would need to extract the contents into a sub-folder which would make the project structure not as convenient.

Another suggestion might be to make an exception if the only file in the folder is a project file?

StevenRasmussen:
Fixes #314 

This PR adds the ability to allow files with particular extensions to exist in the directory where the .msapp file contents are extracted.

### DIFF
--- a/src/PAModel/Utility/DirectoryHelper.cs
+++ b/src/PAModel/Utility/DirectoryHelper.cs
@@ -145,7 +145,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         }
 
         /// <summary>
-        /// Returns true if it's either an empty directory or only includes valid files.
+        /// Returns true if it's either an empty directory or only contains valid files.
         /// </summary>
         /// <returns></returns>
         private bool ValidateSafeToDelete(ErrorContainer errors)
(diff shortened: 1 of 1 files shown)


petrochuk:
Closing old PR

StevenRasmussen:
Curious... was this even reviewed?  Closing a PR simply because it's "old" seems a bit strange.  I'd still like to see this supported.  Are you able to review the PR and comment on it or comment on the feasibility of allowing this functionality?

petrochuk:
We are trying different approach to canvas pack/unpack functionality and older PRs are no longer relevant

PR #320, opened 2021-11-12
Fix "'0x00' is invalid within a JSON string." exception from 'Load' m…

rnwoodhcl:
…ethod when running on .NET 6

Running on .NET introduces a bug when the library calculates file checksums. This causes the exception above under some conditions (file size > some threshold ?).

I have looked into it and it's because the method in this PR doesn't correctly check the return value for Read method that indicates if all requested bytes have been read or not. This results in a byte array being return which has some of the bytes missing and are hence the default null bytes are returned instead. These bytes are then parsed into JsonDocument in CheckSumMaker, and this causes the exception.

The PR fixes this and correctly loops until the whole file contents have been read from the zip entry.

### DIFF
--- a/src/PAModel/Utility/Utilities.cs
+++ b/src/PAModel/Utility/Utilities.cs
@@ -134,9 +134,15 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             using (var s = e.Open())
             {
                 // Non-seekable stream.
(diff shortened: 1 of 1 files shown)


MikeStall (approved):

MikeStall:
Thank you for reporting and providing a PR!

PR #334, opened 2022-01-14
Update to support addition of AllowAccessToGlobals property on component definition

Shpakh-MSFT:
- Add AllowAccessToGlobals property on ComponentDefinitionInfoJson.
- Updates to IRStateHelper to use this property value.


lesaltzm (approved):
:ship:

PR #335, opened 2022-01-19
Fixed AllowGlobalScope update state issue

Shpakh-MSFT:
My previous change to add support for AllowAccessToGlobals property wasn't complete. It failed in roundtripping validation. This change moved the IR property to ComponentManifest instead.


src/PAModel/EditorState/ControlState.cs
```diff
@@ -29,6 +29,8 @@ internal class ControlState
         // even though this can be re-derived from the existence of DynamicProperties
         public bool? HasDynamicProperties { get; set; }
 
+        public bool? AllowAccessToGlobals { get; set; }
```
lesaltzm, line 32: Is this used from ControlState, or was this testing?

src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -336,12 +336,13 @@ private static (ControlInfoJson.Item item, int index) CombineIRAndState(BlockNod
                     StyleName = state.StyleName,
                     ExtensionData = state.ExtensionData,
                     IsGroupControl = state.IsGroupControl,
-                    GroupedControlsKey = state.GroupedControlsKey
+                    GroupedControlsKey = state.GroupedControlsKey,
+                    AllowAccessToGlobals = (state.IsComponentDefinition ?? false) ? (templateState?.ComponentManifest?.AllowAccessToGlobals ?? true) : true,
```
lesaltzm, line 340: this default to true has broken our existing roundtrip tests, looks like most apps don't have this set at all.

src/PAModel/EditorState/ControlState.cs
Shpakh-MSFT: Not needed. Good catch. Thanks.

Shpakh-MSFT: This is now used.

lesaltzm (approved):
🚢

PR #336, opened 2022-01-21
Dynamic Property NullCheck

gesnaaggarwal:
Null check to prevent exceptions generated during yaml to .msapp conversion due to default dynamic properties generated.


src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -503,15 +503,18 @@ private static ControlInfoJson.DynamicPropertyJson GetDynamicPropertyEntry(Contr
             DynamicPropertyState propState = null;
             if (state.DynamicProperties.ToDictionary(prop => prop.PropertyName).TryGetValue(propName, out propState))
             {
-                property.Rule = new ControlInfoJson.RuleEntry()
+                if (propState.Property != null)
```
lesaltzm, line 506: Add a comment about how the DynamicProperties list may contain items without a corresponding property, and those correspond to empty properties, rather than properties with a default value.

lesaltzm (approved):

Issue #325, opened 2021-12-13
Roundtrip validation on unpack failed. Mysterious AppTests folder causing issues


MattTrinder1:
I'm getting the above error trying to unpack the attached (zipped) app.

[cp_suddendeathazurecontinuous_49558_DocumentUri.zip](https://github.com/microsoft/PowerApps-Language-Tooling/files/7705237/cp_suddendeathazurecontinuous_49558_DocumentUri.zip)

MattTrinder1:
This error only occurs if there is a single empty test case present - once the case has had a single step added the error no long happens.

tomjschneider:
This issue has been merged into the Power Platform CLI and will be available in the next release.

PR #340, opened 2022-02-08
Fix roundtrip validation error on unpacking PowerApps with empty test case

NavneetThekkumpat:
Fix for roundtrip validation failure (Git issue #325 )
PowerApps with empty test case/suite results in roundtrip validation failure on unpack.
- It throws hash mismatch error due to diffs in AppTests\\2.json.
- No TestStepsMetadata property is added (but the .msapp Src\EditorState test.json file comes with this property even if the test case is null)
image

With the fix
- TestStepsMetadata property is added with empty test cases too.
- Roundtrip validation is successful since it matches with the original .msapp json file.
- Also updated EmptyTestCase.msapp under src\PAModelTests\Apps considering that any new .msapp created with empty test case will have this TestStepsMetadata property added (not the case earlier I believe and hence to keep the test msapp consistent with current behavior)


lesaltzm (commented):
Please add a description of the issue, and a better title for the PR

lesaltzm (approved):

MikeStall:
Can we add a test case, but keep the existing one working?
By changing an existing test case, we're potentially losing back-compat coverage.

NavneetThekkumpat:
Good point, Sure let me go through the tests and see what can be done.

MikeStall (approved):

PR #341, opened 2022-02-08
Jmstall/yamlpoco

MikeStall:
Introduce a Schema object for strongly typed parameters. 
This defines the schema class in the ALM layer as the source of truth and lets document server include it.

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlPocoSerializer.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlPocoSerializer.cs
@@ -51,7 +51,7 @@ namespace Microsoft.PowerPlatform
             {
                 // Exclude default values.
                 // This provides more stable output as new things are added in the future.
-                return;                
+                return;
             }
             else if (obj is bool valBool)
             {
@@ -88,10 +88,10 @@ namespace Microsoft.PowerPlatform
                         var list = new List<KeyValuePair<string, object>>();
                         foreach (DictionaryEntry kv in dict)
                         {
-                            list.Add(new KeyValuePair<string, object>((string) kv.Key, kv.Value));
+                            list.Add(new KeyValuePair<string, object>((string)kv.Key, kv.Value));
                         }
 
-                        
+
                         yaml.WriteStartObject(propName);
                         WriteCanonicalList(yaml, list);
                         yaml.WriteEndObject();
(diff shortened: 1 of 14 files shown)


src/PAModel/PAConvert/Yaml/YamlPocoSerializer.cs
```diff
@@ -0,0 +1,138 @@
...
+            WriteCanonicalList(yaml, list);
+        }
+
+        // Write a list of properties (such as an object or dictionary) in an ordered way.
+        private static void WriteCanonicalList(YamlWriter yaml, List<KeyValuePair<string, object>> list)
+        {
+            // Critical to sort to preserve a canonical order.
+            list.Sort((kv1, kv2) => kv1.Key.CompareTo(kv2.Key));
+
+            foreach(var kv in list)
```
lesaltzm, line 132: nit: run autoformat on this file

src/PAModel/Schemas/Schema.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PowerPlatform.Formulas.Tools.Schemas
+{
+    // Define Schema.yaml - strong typing for input and output parameters
+    public class ParameterSchema
+    {
+        // What names here?
```
lesaltzm, line 13: nit: remove comment?

src/PAModel/Schemas/Schema.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            // Optional: For entities, specifies the name
+            public string Name { get; set; }
+
+            // Optional: For records, can become recursive
+            public Dictionary<string, Parameter> Fields { get; set; }
+        }
+                
+        // Input parameters (public)
+        public Dictionary<string, Parameter> Parameters { get; set; }
```
lesaltzm, line 43: Do we need all of Parameters/Outputs/Globals to begin with? 
This is an expansion direction, starting out I think we're only focusing on Parameters

lesaltzm (commented):
Please add a more meaningful title and description as well

src/PAModel/Schemas/Schema.cs
MikeStall: It's a schema, so let's lay it out now.

lesaltzm (approved):

PR #343, opened 2022-02-09
Resolve editor state name conflicts during serialization

tomjschneider:
Fixes an issue where controls with the same case-insensitive names can conflict on the file system when generating `editorstate.json` files for the control. This uses the existing name collision handling logic to generate a new filename when needed.

This introduces a case where, when loading in the source files, the `TopParentName` of a control may be set incorrectly, due to the reliance on the filename. This change updates the `ControlState` object to store the `TopParentName` in the appropriate `editorstate.json` during serialization. When loading sources, the `TopParentName` will now be present on the `ControlState` and the filename does not matter.

This change increments the SourceSerializer's version number.

### DIFF
--- a/src/PAModel/EditorState/ControlState.cs
+++ b/src/PAModel/EditorState/ControlState.cs
@@ -1,7 +1,4 @@
-using Microsoft.PowerPlatform.Formulas.Tools.EditorState;
-using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


tomjschneider:
Taking the backwards compatible route for this code, so this PR will not move forward.

PR #344, opened 2022-02-09
Resolve editor state name conflicts during serialization

tomjschneider:
Fixes an issue where controls with the same case-insensitive names can conflict on the file system when generating `editorstate.json` files for the control. This uses the existing name collision handling logic to generate a new filename when needed.

This introduces a case where, when loading in the source files, the `TopParentName` of a control may be set incorrectly, due to the reliance on the filename. This change updates the `ControlState` object to store the `TopParentName` in the appropriate `editorstate.json` during serialization. When loading sources, the `TopParentName` will now be present on the `ControlState` and the filename does not matter.

This change does not increment the SourceSerializer's version number and handles the case where an app was unpacked prior to this change by falling back on the filename of the editor state. Both the new tool using the previous format, and the previous tool using the new format have been tested.

The two new Apps added were created from blank templates to show simple examples of the issues fixed (duplicate screen names and screen/component name collisions.)


src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -369,10 +368,20 @@ private static void LoadSourceFiles(CanvasDocument app, DirectoryReader director
 
                 // Json peer to a .pa file. 
                 var controlExtraData = file.ToObject<Dictionary<string, ControlState>>();
-                var topParentName = file._relativeName.Replace(".editorstate.json", "");
+                var topParentFilename = file._relativeName.Replace(".editorstate.json", "");
                 foreach (var control in controlExtraData)
                 {
-                    control.Value.TopParentName = Utilities.UnEscapeFilename(topParentName);
+                    // START v24 backwards compatibility
```
lesaltzm, line 374: Move to a function? `ApplyV24BackCompat(control)`?

src/PAModel/EditorState/ControlState.cs
```diff
@@ -14,12 +11,11 @@ internal class ControlState
     {
         public string Name { get; set; }
 
-        [JsonIgnore]
```
lesaltzm, line 17: Are we writing this for every control in the editor state file? Is there a way we could only write it once?

src/PAModel/Serializers/SourceSerializer.cs
tomjschneider: Good idea, I'll do that.

src/PAModel/EditorState/ControlState.cs
tomjschneider: I believe that yes, it is being written as a property for every control. Let me dig on this a bit more to see what we can do here.

tomjschneider: I've reworked this so that we're not storing it on each control but instead denoting the top parent control only. To avoid adding additional time complexity when reading in the editor state during deserialization, I've also used a well-known name for the dictionary key of the top parent control. This key was otherwise not being used, and lets us immediately find the top parent and use its name.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -370,6 +370,11 @@ private static void LoadSourceFiles(CanvasDocument app, DirectoryReader director
                 // Json peer to a .pa file. 
                 var controlExtraData = file.ToObject<Dictionary<string, ControlState>>();
                 var topParentName = file._relativeName.Replace(".editorstate.json", "");
+
+                // Check for the designated parent control. Fall back on the file name.
+                if (controlExtraData.ContainsKey(TopParentObjectKey) && (controlExtraData[TopParentObjectKey].IsTopParent ?? false))
```
lesaltzm, line 375: So, why do we need this TopParentObjectKey? 
We could just serialize as is, and then the deserialize retrieval could be 
`controlExtraData.FirstOrDefault(kvp => kvp.Value.IsTopParent)) ?? topParentName`

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -886,20 +891,26 @@ private static void RestoreViewName(DataSourceEntry ds, string dataSetName)
...
+            // parent object a unique key so it can be retrieved when loaded.
+            var controls = new Dictionary<string, ControlState>();
             foreach (var item in app._editorStateStore.GetControlsWithTopParent(topParentname ?? controlName))
             {
-                extraData.Add(item.Name, item);
+                if (item.IsTopParent ?? false)
+                    controls.Add(TopParentObjectKey, item);
+                else
+                    // Preface with an underscore to prevent unlikely collision with top parent
+                    controls.Add($"_{item.Name}", item);
```
lesaltzm, line 903: This seems a bit hacky. We shouldn't be putting names into unpacked files that aren't obviously meaningful if we can avoid it.

src/PAModel/Serializers/SourceSerializer.cs
tomjschneider: Restructured this to use a specific object model for editor state files.

src/PAModel/EditorState/EditorStateFile.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerPlatform.Formulas.Tools.EditorState
+{
+    /// <summary>
+    /// Represents the `editorstate.json` file created to contain ControlStates,
+    /// which have studio state content that don't wind up in the IR.
+    /// </summary>
+    internal class EditorStateFile
```
lesaltzm, line 12: Naming nit: maybe `TopParentTreeState`? It's not a file, the file ending could be confusing

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -886,20 +897,24 @@ private static void RestoreViewName(DataSourceEntry ds, string dataSetName)
...
-            var extraData = new Dictionary<string, ControlState>();
-            foreach (var item in app._editorStateStore.GetControlsWithTopParent(topParentname ?? controlName))
+            // For TestSuite controls, only the top parent control has an editor state created.
+            // For other control types, create an editor state.
+            if (string.IsNullOrEmpty(topParentName))
             {
-                extraData.Add(item.Name, item);
-            }
+                string editorStateFilename = $"{newControlName}.editorstate.json";
+                EditorStateFile editorStateFile = new EditorStateFile
```
lesaltzm, line 905: Consider moving this constructor after the dictionary is populated. We aren't actually using immutabe data structures here, but it's still generally good practice to not insert into a property of another object directly.

src/PAModel/EditorState/EditorStateFile.cs
tomjschneider: Agreed on the "File". I don't want to misrepresent this object as the top parent of the tree (since that control is actually contained in `ControlStates`). Thoughts on `EditorControlState`? If it still seems unclear I'm fine to go with `TopParentTreeState` as well.

tomjschneider: I went with `ControlTreeState`. Let me know what you think and I'll adjust if needed.

Issue #300, opened 2021-08-05
Error PA3011: Roundtrip validation on unpack failed


Chris-WP:
We are experiencing a new unpack issue with the latest solution package we have developed.

The error is:
```Warning PA2001: Checksum mismatch. File Controls\1322.json checksum does not match on extract
Error   PA3011: Roundtrip validation on unpack failed. This is always a bug, please file an issue at https://github.com/microsoft/PowerApps-Language-Tooling
Error   PA3011: Roundtrip validation on unpack failed. This is always a bug, please file an issue at https://github.com/microsoft/PowerApps-Language-Tooling
2 errors, 1 warnings.```

We are experiencing this issue with both the Powert Platform CLI v1.8.6 and the latest code base in the Git repository (i.e. up to commit 6a884a9.

@MikeStall: I'll email you the solution file that contains the .msapp file.

**Comment by mikefactorial** [comment]
Just ran into the same 

```
UnPack from D:\a\1\s\coe-starter-kit\CenterofExcellenceCoreComponents\SolutionPackage\CanvasApps\coe_environmentrequestappadmin_58e3b_DocumentUri.msapp-->D:\a\1\s\coe-starter-kit\CenterofExcellenceCoreComponents\SolutionPackage\CanvasApps\coe_environmentrequestappadmin_58e3b_DocumentUri_msapp_src
  source format version: 0.24
Error   PA3011: Roundtrip validation on unpack failed. This is always a bug, please file an issue at https://github.com/microsoft/PowerApps-Language-Tooling
Error   PA3011: Roundtrip validation on unpack failed. This is always a bug, please file an issue at https://github.com/microsoft/PowerApps-Language-Tooling
Microsoft PowerPlatform CLI
Version: 1.8.5+g2039663

Error: There were errors.

```
Solution
[CenterofExcellenceCoreComponents_2_73.zip](https://github.com/microsoft/PowerApps-Language-Tooling/files/6956292/CenterofExcellenceCoreComponents_2_73.zip)
 is attached.

bruhaha1313:
Same. I have nothing else useful to add because the log only reports a generic error.

2021-08-10T14:01:07.4307100Z ##[section]Starting: unpack msapp files
2021-08-10T14:01:07.4410032Z ==============================================================================
2021-08-10T14:01:07.4410425Z Task         : PowerShell
2021-08-10T14:01:07.4410777Z Description  : Run a PowerShell script on Linux, macOS, or Windows
[...]

PR #348, opened 2022-02-18
Handling dynamic properties default value in defaultValueTransform

NavneetThekkumpat:
Roundtrip failed while unpacking the .msapp with fluidGrid container #300
Problem seems to be
- hash mismatch because of dynamic property with null values included in control jsons.
image

Solution
-By fixing the way dynamic property with null value is handled in src\PAModel\SourceTransforms\DefaultValuesTransform.cs


NavneetThekkumpat:
> Not sure how doable it is but might want to look into if this use case could be covered by a unit test.

Good point. Added unit test covering the same.

tomjschneider (approved):
Looks good - thanks for taking the time to add a unit test!

landonarogers (approved):

PR #351, opened 2022-02-24
Handling pcfTemplate data in templates.json while roundtripping

NavneetThekkumpat:
Problem:
Unpacking references\templates.json (\src\PAModel\Schemas\TemplatesJson.cs) currently
- UsedTemplates and ComponentTemplates are handled/recognized now.
- PcfTemplates are not handled/recognized and hence ignored while unpacking.

Solution:
- Extract the pcfTemplates values onto another file(pkgs/pcfConversions.json) while unpacking templates.json
- This data can be added back to the templates.json while repacking.

Other Possible Solution:
- ControlTemplate.json file which contains control template info. We could add a top-level section to that called "PcfTemplateConversions" which stores this info and is used to re-populate it on packing.


tomjschneider (changes requested):
Mostly minor changes in my feedback - I think splitting this out to a new file makes sense, the existing `ControlTemplate.json` uses a specific format. 

Would also recommend generating and adding a sample app to the roundtrip tests for testing purposes.

tomjschneider (approved):
Navaneeth is in training all week, so addressed feedback myself. Would like someone else to take a look before I actually complete the PR.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -321,10 +321,24 @@ private static void LoadTemplateFiles(ErrorContainer errors, CanvasDocument app,
...
             }
 
+            List<PcfTemplateJson> pcfTemplateConversionsList = null;
+            string pcfTemplatePath = Path.Combine(packagesPath, PCFConversionFileName);
+            if (File.Exists(pcfTemplatePath))
+            {
+                pcfTemplateConversionsList = new List<PcfTemplateJson>();
+
+                DirectoryReader.Entry file = new DirectoryReader.Entry(pcfTemplatePath);
+                var pcfTemplateConversions = file.ToObject<PcfTemplateJson[]>();
```
MikeStall, line 331: >pcfTemplateConversions [](http://example.com/codeflow?start=20&length=22)

What's the difference between this and pcfTemplateConversionsList?

NavneetThekkumpat: Thanks for catching this. Was handling the array differently earlier while working, now we don't need pcfTemplateConversionsList.

MikeStall (approved):
:shipit:

MikeStall (approved):
:shipit:

PR #353, opened 2022-03-03
Users/landonarogers/bug fix 3366

landonarogers:
Problem:
Right now when the PASopa tool hits a roundtripping error it writes it out to two static files: "a1.json" and "b1.json". This needs to be fixed for two reasons:

- It assumed the files exist. If they don't it errors out.
- If there are multiple errors it will constantly overwrite the files and you'll only get the "last" roundtripping error it hits.

Solution:

- Creates 'diffFiles' directory in bin/Debug/PASopa if it doesn't exist.
- Creates new files if they don't exist.
- Will not overwrite files.
- Dynamically creates files 'A' and 'B' for each file with diffs, each containing the specific file's name.


tomjschneider (approved):
Looks good.

NavneetThekkumpat (approved):

Issue #308, opened 2021-09-10
Error   PA3005: Validation error: Connection does not have a corresponding data source.


famastefano:
Hello,
just downloaded the extension for VS and tried to unpack an app but this error shows up:
`Error PA3005: Validation error: Connection 'Docs_1' does not have a corresponding data source.`

The fact is that I can't see **Docs_1** anywhere in the Data Sources, maybe it was removed, I don't know.

```
Microsoft PowerPlatform CLI
Versione: 1.9.4+gbda9f20
```

Bregor-Mar-Rian:
Same issue here. It fails with the same type of error.
The app had a data connection that was removed at some point in a standard way.

For some of our other apps (where we also deleted certain old data connections), same [unpack](https://docs.microsoft.com/en-us/powerapps/developer/data-platform/cli/reference/canvas-command#:~:text=msapp%20HelloWorld.msapp-,unpack,-Unpacks%20the%20.msapp) seems to work fine.

The only place we can see a reference to the old connection in is the exported package (zip) version:
Microsoft.PowerApps → apps → [app_unique_number] → [app_unique_number].json → appDefinitionTemplate → properties → connectionReferences → dataSources

Currently, this is a show-stopper for our app that needs source control the most.
How can we fix this error ?

MikeStall:
Can you email your msapp to jmstall at microsoft dot com

landonarogers:
The latest version of the PAC CLI (1.13.4) has been released which includes support for this issue. If you are still experiencing roundtripping errors on the latest version of the CLI, please open a new Issue in the repo.

PR #354, opened 2022-03-07
Issue 308: Removing Validation For Missing Data Source

landonarogers:
Problem:
Issue originally documented here: https://github.com/microsoft/PowerApps-Language-Tooling/issues/308 . When Properties.json is missing a table for a specific corresponding Data Source, it will cause this issue and refuse to unpack. However, it does not seem necessary for this validation error to be thrown because there is no roundtrip error otherwise.

Solution:
Remove the code that throws this validation error.


tomjschneider (approved):
Looks good.

PR #356, opened 2022-03-07
Bug: Unsuccessful Unpack Corrupts Existing Files

landonarogers:
Problem:
When a roundtrip failure occurs, the previously unpacked files were removed causing potential data loss.

Solution:
Unpack in a temporary folder. If unsuccessful, remove temporary folder. If successful, move contents to destination folder, then remove temporary folder.

### DIFF
--- a/src/PASopa/Program.cs
+++ b/src/PASopa/Program.cs
@@ -126,6 +126,10 @@ namespace PASopa
                     outDir = args[2];
                 }
 
+
+                // Use this directory for temporarily unpacking files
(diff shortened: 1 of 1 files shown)


tomjschneider:
I'm not sure this will solve the entire problem. This change is in the PASopa CLI within this project, which isn't typically being run. Most people are interacting with this through the PAC CLI, which only uses the class library. Let's chat about this a little more offline and what scenario we're trying to fix.

PR #361, opened 2022-03-15
Bring ParameterSchema in line with PowerFx FormulaTypeSchema

lesaltzm:
Unifying definitions with https://github.com/microsoft/Power-Fx/pull/215

Eventually, we'll update this tool to depend on PowerFx and pull the definition directly.


tomjschneider (approved):
Looks good to me. I double checked [the Power-Fx repo](https://github.com/microsoft/Power-Fx/blob/main/src/libraries/Microsoft.PowerFx.Core/Public/Types/Serialization/FormulaTypeSchema.cs) and everything that should match appears to.

PR #363, opened 2022-03-22
Github Issue 305 (Controls in the fx.yaml not canonical)

landonarogers:
Problem:
Controls cause noise by being listed in an inconsistent order, when ZIndexes are the same value, and has to determine order by itself.

Solution:
Ensure that the controls in the fx.yaml are being listed in a canonical order when ZIndexes are the same value.


lesaltzm:
Is alphabetizing the controls the right solution here? Imo, ideally, the order of controls in the .fx.yaml files should match the order of controls in the Tree View in studio, which is stable. What is studio doing to maintain that order?

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -362,6 +362,12 @@ private static void GetLogoFile(this CanvasDocument app)
             }
         }
 
+        // Stores an .fx.yaml file and it's directory bool value
+        public struct fxYamlGroup{
```
lesaltzm, line 366: nit: struct nanes start with capital letter

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -362,6 +362,12 @@ private static void GetLogoFile(this CanvasDocument app)
...
+            public Microsoft.PowerPlatform.Formulas.Tools.DirectoryReader.Entry file;
```
lesaltzm, line 367: Does this need to be a struct, or should it just be a tuple used local to this function?

i.e. `List<(Entry file, bool canSearch)>`

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -411,9 +427,25 @@ private static void LoadSourceFiles(CanvasDocument app, DirectoryReader director
...
+                // Add fxYamlGroup object to the list
+                fxYamlList.Add(currFxYamlGroup);
             }
 
+            // Sort the list of fxYaml files alphabetically
+            fxYamlList.OrderBy((fxYamlGroup) => fxYamlGroup.file._relativeName);
+
+            // Add control for every .fx.yaml file
+            fxYamlList.ForEach(delegate(fxYamlGroup group ) {
+            AddControl(app, group.file._relativeName, group.canSearchSubdirs, group.file.GetContents(), errors);
```
lesaltzm, line 446: I don't think this is actually solving the issue mentioned in the PR re the order of the controls within the file. This looks like it's sorting at the file level, which is not going to have any effect since the filesystem, and zip files are unordered.

landonarogers: I want to order it by what the control name is going to be, but it doesn't seem like I have any way to do that unless I want to invoke the parser in this section of the code. I may have to move to another section of code unless it's okay to invoke the parser up here, just to grab the controlIR name and order alphabetically

landonarogers:
> Is alphabetizing the controls the right solution here? Imo, ideally, the order of controls in the .fx.yaml files should match the order of controls in the Tree View in studio, which is stable. What is studio doing to maintain that order?

It was mentioned that ordering it alphabetically will stop it from mysteriously changing between each minor change. Do you mean that we should make the order in the Tree View in studio alphabetical so that the .fx.yaml files naturally match?

landonarogers:
> Is alphabetizing the controls the right solution here? Imo, ideally, the order of controls in the .fx.yaml files should match the order of controls in the Tree View in studio, which is stable. What is studio doing to maintain that order?

I was trying to add both sections of controls that are being added here. It seemed to me that, since they are getting added separately, they are being added in an out of sync order:

`            foreach (var file in EnumerateComponentDirs(directory, "*.fx.yaml"))
            {
                // Add tuple to the list
                fxYamlList.Add((file, true));
            }

            foreach (var file in EnumerateComponentDirs(directory, "*.json"))
            {
                var componentTemplate = file.ToObject<CombinedTemplateState>();
                app._templateStore.AddTemplate(componentTemplate.ComponentManifest.Name, componentTemplate);
            }

            foreach (var file in directory.EnumerateFiles(CodeDir, "*.fx.yaml", searchSubdirectories: false))
            {
                // Add tuple to the list
                fxYamlList.Add((file, false));
            }`

tomjschneider (commented):
Code looks good, recommend adding a unit test to `SourceDecoderTests` with a simple test app which ensures the expected ordering.

tomjschneider (approved):
Looks good - thanks for adding the unit test!

PR #364, opened 2022-03-22
Draft: Github issue 191 (validation in themes.json)

landonarogers:
Problem:
When the Themes.json file is changed to contain an empty string for the default script, it causes the program to crash without an error.

Solution:
Print out a new error that lets the user know there are empty strings for the default scripts in the Themes.json file.

### DIFF
--- a/src/PAModel/Themes/Theme.cs
+++ b/src/PAModel/Themes/Theme.cs
@@ -48,7 +48,8 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 var styleName = style.name;
                 foreach (var prop in style.propertyValuesMap)
                 {
-                    if(prop.value == ""){
+                    if (prop.value == "")
+                    {
                         errors.ValidationError($"Themes.json cannot contain a default script of empty string.");
                     }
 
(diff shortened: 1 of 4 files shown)


src/PAModel/Themes/Theme.cs
```diff
@@ -46,6 +46,10 @@ private void LoadTheme(CustomThemeJson theme)
                 var styleName = style.name;
                 foreach (var prop in style.propertyValuesMap)
                 {
+                    if(prop.value == ""){
```
lesaltzm, line 49: Should the ErrorContainer be passed down to this? ArgumentException doesn't seem like the right way to handle this, it's possible this should only be a warning.

lesaltzm: Basically, I think there might be some apps out there with this condition that are actively in use, since Doc Server doesn't enforce it strongly. It only crashes in some scenarios, I think if the default script is used, or something like that.

landonarogers: Okay; I'm currently trying to connect an ErrorContainer to this. Where are the ErrorContainers typically in the lifecycle of the program, so that I can pass it down the code towards this point?

lesaltzm: There's one created at the start of each operation in CanvasDocument, used for the lifetime of that operation. They should be available in all the places we're calling the Theme constructor

lesaltzm, line 49: nit: space after if, and opening brace of the block on new line.

landonarogers: Okay, thank you! I've connected the ErrorContainer now :) Now testing

landonarogers:
Could no longer reproduce the original issue causing the crash. Closing PR

PR #365, opened 2022-03-23
Adding new field to correctly determine whether LocalDatabaseReferences was empty or not and avoid hash mismatch

akshar24:
1) Adding a new field calledto separately track whether LocalDatabaseReferences in Properties.json was empty or not.
2) Changing the code where "LocalDatabaseReferencesAsEmpty" was incorrectly used to check whether "LocalDatabaseReferences" was empty or not
3) Handling backward compatibility with the new field
4) Skipping Data Source References with empty Data Sources


src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -299,16 +299,18 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
                     app._properties.LocalConnectionReferences = null;
                 }
 
-                if (!string.IsNullOrEmpty(app._properties.LocalDatabaseReferences))
+                IDictionary<String, LocalDatabaseReferenceJson> dsrs = null;
```
lesaltzm, line 302: nit: string not String

src/PAModel/Utility/Utilities.cs
```diff
@@ -453,5 +453,12 @@ public static void VerifyDirectoryExists(ErrorContainer errors, string fullpath)
                 errors.BadParameter($"Directory not found: {fullpath}");
             }
         }
+
+        public static bool IsLocalDatabaseReferencesEmpty(this Entropy entropy)
+        {
+            return (entropy.WasLocalDatabaseReferencesEmpty != null) ?
+                   (bool)entropy.WasLocalDatabaseReferencesEmpty:
```
lesaltzm, line 460: For nullable properties, use `.Value` to retrieve the value, and `.HasValue` to test instead of comparing to null.

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -299,16 +299,18 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
...
+                if (!string.IsNullOrEmpty(app._properties.LocalDatabaseReferences) && (dsrs = Utilities.JsonParse<IDictionary<String, LocalDatabaseReferenceJson>>(app._properties.LocalDatabaseReferences)).Count > 0)
```
lesaltzm, line 303: Assignment that can be short circuited is a little hard to read. Could this be refactored to not have the assignment part of the condition?

src/PAModel/Utility/Utilities.cs
```diff
@@ -453,5 +453,12 @@ public static void VerifyDirectoryExists(ErrorContainer errors, string fullpath)
                 errors.BadParameter($"Directory not found: {fullpath}");
             }
         }
+
+        public static bool IsLocalDatabaseReferencesEmpty(this Entropy entropy)
```
lesaltzm, line 457: This doesn't need to be an extension method, you can just add the method on Entropy.cs

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -791,6 +791,9 @@ private static void LoadDataSources(CanvasDocument app, DirectoryReader director
                     localDatabaseReferenceJson.dataSources.Add(tableDef.EntityName, tableDef.LocalReferenceDSJson);
                 }
             }
+            //Skip the data source references with empty data sources. This avoids data source reference with empty json for data sources
```
lesaltzm, line 794: nit: Space after start of comment

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -791,6 +791,9 @@ private static void LoadDataSources(CanvasDocument app, DirectoryReader director
...
+            app._dataSourceReferences = dataSourceReferences.Where(dataSourceReferenceEntry => dataSourceReferenceEntry.Value.dataSources.Count > 0)
+                                                            .ToDictionary(dataSourceReferenceEntry => dataSourceReferenceEntry.Key, dataSourceReferenceEntry => dataSourceReferenceEntry.Value);
```
lesaltzm, line 796: `dataSourceReferences.Where(dataSourceReferenceEntry => dataSourceReferenceEntry.Value.dataSources.Count > 0)` is an IEnumerable<KeyValuePair<string, LDRJson>> where we've already ensured uniqueness of keys

Instead of using .ToDictionary, just pass it to a dictionary constructor. Avoids extra allocations for LINQ

src/PAModel/Serializers/MsAppSerializer.cs
akshar24: Changed String to string

src/PAModel/Utility/Utilities.cs
akshar24: Moved this method to Entropy.cs. Refactored this condition into null coalescing

src/PAModel/Serializers/MsAppSerializer.cs
akshar24: Refactored this block to remove assignment from the condition. Need to check for empty string or null before performing Json parsing.

src/PAModel/Utility/Utilities.cs
akshar24: Moved the method to Entropy.cs

src/PAModel/Serializers/SourceSerializer.cs
akshar24: Thanks for catching this. Added space

src/PAModel/Serializers/SourceSerializer.cs
lesaltzm: why not ` app._dataSourceReferences = new Dictionary<.., .. > (dataSourceReferences.Where(dataSourceReferenceEntry => dataSourceReferenceEntry.Value.dataSources.Count > 0))`

akshar24: I may be misunderstanding it here but i couldn't find overload for Dictionary constructor that takes in IEnumerable<KeyValuePair<...>>. I changed the code to avoid extra allocations of KeyValuePair objects. I need to filter out keyvaluepairs with empty data sources from dataSourceReferences local var and move those to app.dataSourceReferences.

lesaltzm: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.-ctor?view=netcore-3.1#system-collections-generic-dictionary-2-ctor(system-collections-generic-ienumerable((system-collections-generic-keyvaluepair((-0-1)))))

akshar24: Target framework for Microsoft.PowerPlatform.Formulas.Tools is netstandard2.0. "Applies To" table for this overloaded constructor only mentions .NET standard 2.1. I only get 6 overloaded constructor options and I get compiler error when I try to pass in the Ienumerable of keyvalue pairs as an argument to dictionary constructor. Not sure if something is wrong with my local setup

lesaltzm: Ah, nope. My bad! I didn't check that it wasn't in Standard 2.0. Feel free to ignore this.

akshar24: No worries! It's good to know that this exists. Didn't know about this overloaded constructor. Would have used Linq in other places as well. Thanks for making me aware of this.

src/PAModel/Entropy.cs
```diff
@@ -71,11 +71,20 @@ internal class PropertyEntropy
 
         public PropertyEntropy VolatileProperties { get; set; }
 
-        // Sometimes, an empty LocalDatabaseReferences is serialized as "" or { }.
-        // Need to track which so we can round-trip.
-        // if true: "".   Else,  "{}". 
+        /// <summary>
```
NavneetThekkumpat, line 74: nit : I believe we generally use "//" for single line comments rather than "///" int his repo. Would be great if we could stick to that.

NavneetThekkumpat:
Thanks for adding the unit test. looks good

PR #368, opened 2022-03-24
Add DisplayNameKeyProperty to PCF Schemas

jack-work:
A new property, TemplateMetadataJson, is being serialized to the DynamicControlDefinitionJson within TemplateMetadataJson.  This change must be reflected by both the PcfControl class and, because DynamicControlDefinitionJson is double-serialized, to PcfControlDoubleSerializedClass.  This change facilitates the reflection.


yiwl0612 (approved):

src/PAModel/Schemas/PcfControls/PcfControl.cs
```diff
@@ -16,6 +16,7 @@ internal class PcfControl
         public string Version { get; set; }
 
         public string ControlNamespace { get; set; }
+        public string DisplayNameKey { get; set; }
```
lesaltzm, line 19: This is another case where making this repo the source of truth for POCOs will prevent issues.

lesaltzm: I'm curious, why was this not caught by ExtensionData?

jack-work: Absolutely.  There is a notable and well known caveat here that I will note for posterity.  Because the DynamicControlDefinitionJson object is doubly serialized, and is deserialized by Studio using string-based property access, the source of truth is baked into the code itself.  Moving this particular schema to this repository necessitates the use of an actual schema class to begin with.  The burden of updating PA-Client is greater than in other schematic cases.

jack-work: Anyways, if there is no problem here, then I will resolve.

lesaltzm (approved):

lesaltzm (approved):

lesaltzm (approved):

PR #371, opened 2022-03-25
Need more specific error message for "Roundtrip validation on unpack failed"

landonarogers:
Problem:
Need to be able to narrow these down to more specific issues for tracking & bug filing. Logs need to give better info when this error appears. Right now, the message is too vague; it appears to say that it's found a specific bug, however the message actually indicates only that some bug has been encountered.

Solution:

- Explain that this is not a specific bug, rather just an indicator of a bug.
- Print out all JSON Property names that are mismatched, and if they were created, deleted, or changed.


landonarogers:
Adding unit testing

src/PAModel/MsAppTest.cs
```diff
@@ -179,10 +180,12 @@ public static bool StressTest(string pathToMsApp)
                         return false;
                     }
 
+                    ErrorContainer errorContainer = new ErrorContainer();
```
tomjschneider, line 183: It looks like an `ErrorContainer` is being created just above this (line 163). Is there any reason we can't use that for the Compare call?

src/PAModel/MsAppTest.cs
```diff
@@ -226,12 +229,14 @@ public static bool Compare(string pathToZip1, string pathToZip2, TextWriter log)
...
             // Provide a comparison that can be very specific about what the difference is.
             var comp = new Dictionary<string, byte[]>();
-            DebugChecksum(pathToZip1, log, comp, true);
-            DebugChecksum(pathToZip2, log, comp, false);
 
+            compareChecksums(pathToZip1, log, comp, true, errorContainer);
+            compareChecksums(pathToZip2, log, comp, false, errorContainer);
+
+#if DEBUG
             foreach (var kv in comp) // Remaining entries are errors.
```
tomjschneider, line 240: Do we still need this statement? What is this doing different that wouldn't be covered by the property comparison?

src/PAModel/MsAppTest.cs
```diff
@@ -269,56 +274,82 @@ public static void DebugChecksum(string pathToZip, TextWriter log, Dictionary<st
...
+                                Console.WriteLine("FAIL: 2nd has added file: " + entry.FullName);
                             }
                         }
                     }
                 }
             }
+
+        }
+
+        public static void checkPropertyMismatch(ZipArchiveEntry entry, byte[] otherContents, byte[] key, ErrorContainer errorContainer){
```
tomjschneider, line 308: nit: Method name casing

src/PAModel/MsAppTest.cs
```diff
@@ -269,56 +274,82 @@ public static void DebugChecksum(string pathToZip, TextWriter log, Dictionary<st
...
+            foreach (var currentProperty2 in json2.EnumerateObject())
+            {
+                // If current property from second json file does not exist in the first file
+                if(!json1.TryGetProperty(currentProperty2.Name, out JsonElement value1)){
+                    errorContainer.JSONMismatch(currentProperty2.Name + ": Property Added");
+                }
+            }
+        }
+
+        public static void debugMismatch(ZipArchiveEntry entry, byte[] otherContents, byte[] key, string normFormDir)
```
tomjschneider, line 342: nit: Method name casing

src/PAModel/MsAppTest.cs
```diff
@@ -269,56 +274,82 @@ public static void DebugChecksum(string pathToZip, TextWriter log, Dictionary<st
...
+            JsonElement json2 = JsonDocument.Parse(otherContents).RootElement;
+
+            // Check each property and value in json1 to see if each exists and is equal to json2
+            foreach (var currentProperty1 in json1.EnumerateObject())
+            {
+                // If current property from first json file also exists in the second file
+                if(json2.TryGetProperty(currentProperty1.Name, out JsonElement value2)){
+
+                    // If current property value from first json file is not the same as in second
+                    if(!currentProperty1.Value.Equals(value2)){
```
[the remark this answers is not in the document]
jt000: I noticed that as well. You may need to use "Deep Equal" to check for the subkey, but this will only output that "Key2" has changed. You may need to traverse into JObject types to identify that "SubKey1" has changed.

Consider using the [JsonTextReader](https://www.newtonsoft.com/json/help/html/t_newtonsoft_json_jsontextreader.htm) which will enumerate all the values in a file. From here you can use the [Path](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonReader_Path.htm) property to output the JSON path that might differ (i.e `Key2/SubKey1`)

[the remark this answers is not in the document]
landonarogers: attempting recursive function to iterate through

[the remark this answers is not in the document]
landonarogers: I believe I've resolved this now. Can you take a look? @tomjschneider @jt000

src/PAModel/MsAppTest.cs
```diff
@@ -211,41 +211,54 @@ public static bool StressTest(string pathToMsApp)
...
 
-            foreach (var kv in comp) // Remaining entries are errors.
+            CompareChecksums(pathToZip1, log, comp, true, errorContainer);
+            CompareChecksums(pathToZip2, log, comp, false, errorContainer);
+
+            return false;
+        }
+
+        // Overload with ErrorContainer
+        public static bool Compare(string pathToZip1, string pathToZip2, TextWriter log, ErrorContainer errorContainer)
```
tomjschneider, line 238: For the overload, we want to avoid code duplication where possible. I meant something more along the lines of creating a new `Compare` with an `ErrorContainer` (which you have done) and then changing the previous `Compare` function to look something like:

```
public static bool Compare(string pathToZip1, string pathToZip2, TextWriter log)
{
   ErrorContainer errorContainer = new ErrorContainer();
   return Compare(pathToZip1, pathToZip2, log, errorContainer);
}
```

Now your meaningful logic is contained in a single location, but you can continue to call the old method definition and it will be handled as expected.

tomjschneider (commented):
Would you mind attaching a screenshot to the PR showing an example of what the output look like now when an error is hit?

src/PAModel/MsAppTest.cs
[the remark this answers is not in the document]
tomjschneider: Let's chat about this a bit more in standup, I'm not sure my understanding of what we want to report on is entirely in sync.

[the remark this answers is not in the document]
landonarogers: removed recursive

src/PAModel/CanvasDocument.cs
```diff
@@ -198,7 +198,7 @@ public ErrorContainer SaveToSources(string pathToSourceDirectory, string verifyO
                         return errors2;
```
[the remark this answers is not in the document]
NavneetThekkumpat: Never mind, you can ignore this. No changes needed I believe.

src/PAModelTests/ErrorTests.cs
```diff
@@ -55,5 +57,25 @@ public void OpenMissingSources()
...
+        {
+            (var doc, var errors) = CanvasDocument.LoadFromMsapp(PathJSONMismatchDir);
+            Assert.IsTrue(errors.HasErrors);
+            Assert.IsNull(doc);
+
+            bool containsJSONMismatch = false;
+
+            foreach (var err in errors)
+            {
+                if (err.Code == (ErrorCode)3013)
```
NavneetThekkumpat, line 72: nit:  can we use the enum value here instead of 3013 in the check.( Errorcode.JSONMismatch ).

src/PAModelTests/ErrorTests.cs
```diff
@@ -55,5 +58,148 @@ public void OpenMissingSources()
             Assert.IsTrue(errors.HasErrors);
             Assert.IsNull(doc);
         }
+
+        [TestMethod]
```
cherie-msft, line 62: You may want to consider using inline data to make these unit tests data driven
https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test#add-more-tests

src/PAModelTests/ErrorTests.cs
```diff
@@ -55,5 +58,148 @@ public void OpenMissingSources()
...
+
+            // Assume no mismatch
+            bool JSONPropertyAdded = false;
+            bool JSONPropertyRemoved = false;
+            bool JSONValueChanged = false;
+
+            // For every error received, check JSON Mismatches
+            foreach (var error in errorContainer)
+            {
+                if (error.Code == ErrorCode.JSONPropertyAdded)
```
cherie-msft, line 89: Do we have tests to validate that the message we output is correct? Like the nested paths are output in the correct way?

src/PAModelTests/ErrorTests.cs
landonarogers: Added!

src/PAModelTests/ErrorTests.cs
landonarogers: Added!

landonarogers: Made new changes we discussed!

NavneetThekkumpat (approved):

src/PAModelTests/Text/addedOutput.txt
```diff
@@ -0,0 +1,5 @@
+Error   PA3014: Property Added: Noun.Adjectives[1].Colors[0].Primary
```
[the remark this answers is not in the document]
landonarogers: Added!

tomjschneider (approved):
Few nits, but looks good to me.

PR #373, opened 2022-03-28
Making YamlWriter a public class

tomjschneider:
Making `YamlWriter` a public class so it can be utilized by other projects looking to conform to the Power App specific flavor of YAML.

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlWriter.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlWriter.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.IO;
-using System.Text;
 
 namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
 {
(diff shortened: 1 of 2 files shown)


NavneetThekkumpat (approved):

src/PAModel/PAConvert/Yaml/YamlWriter.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
     /// Notably, property values are always either multi-line escaped or
     /// prefixed with '=' to block yaml from treating it as a yaml expression.
     /// </summary>
-    internal class YamlWriter
+    public class YamlWriter
     {
```
lesaltzm, line 16: Double checking, is there anything in here that becomes public that we don’t need to be?

tomjschneider: The class itself isn't tied to anything else and everything else in the same namespace is marked `internal`. I want everything that is already public on it to be available, so I think we're all good here.

tomjschneider: Ah you know what - `NormalizeNewlines` is only used within this class. Let me mark that as private, it doesn't need to be exposed.

tomjschneider: Okay, changed that to private and removed the static designation as it didn't seem necessary.

lesaltzm (approved):

PR #374, opened 2022-03-28
StyleCop enforcement on build

NavneetThekkumpat:
As more and more PRs are made in PALT repo, we need a solution to improve the PR process. This task enforces stylecop code analyzers to address this problem.

StyleCop Analyzer documentation https://github.com/DotNetAnalyzers/StyleCopAnalyzers/tree/master/documentation .

Enforcing only certain rules for now including Readability Rules and few Naming Rules


src/PAModel/ControlTemplates/GlobalTemplates.cs
```diff
@@ -79,8 +79,8 @@ private static ControlTemplate CreateScreenTemplate(AppType type)
             template.InputDefaults.Add("Fill", "RGBA(255, 255, 255, 1)");
             template.InputDefaults.Add("Height", "Max(App.Height, App.MinScreenHeight)");
             template.InputDefaults.Add("Width", "Max(App.Width, App.MinScreenWidth)");
-            template.InputDefaults.Add("Size", "1 + CountRows(App.SizeBreakpoints) - CountIf(App.SizeBreakpoints, Value >= Self.Width)");
-            template.InputDefaults.Add("Orientation", "If(Self.Width < Self.Height, Layout.Vertical, Layout.Horizontal)");
+            template.InputDefaults.Add("Size", "1 + CountRows(App.SizeBreakpoints) - Countif (App.SizeBreakpoints, Value >= Self.Width)");
+            template.InputDefaults.Add("Orientation", "if(Self.Width < Self.Height, Layout.Vertical, Layout.Horizontal)");
```
jt000, line 83: I'm not sure if this will continue to work... Are PowerFX expressions case-insensitive?

src/PAModel/Schemas/PropertyDataType.cs
```diff
@@ -7,25 +7,25 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Schemas
...
-        Hyperlink,
-        Currency,
-        Image,
-        Color,
-        Enum,
-        Media,
-        Guid,
-        Screen,
-        _Lim
+    Invalid,
```
jt000, line 10: Indenting seems correct here. Shouldn't anything within curly braces be indented?

NavneetThekkumpat: I was wondering the same, but <Rule Id="SA1137" Action="None" /> 
Was throwing this error before.

src/PAModel/ControlTemplates/GlobalTemplates.cs
NavneetThekkumpat: Good point, thanks for catching, restoring these. Changed with other files automatically.

lesaltzm (approved):

PR #376, opened 2022-03-30
Handling connection instance id and avoiding noisy diffs

NavneetThekkumpat:
Problem : Noisy diffs because of connection instance ids

Solution:
- Extracting connection instance ids from connections.json to entropy.json from properties.json while unpacking
- Making sure to add it back to properties while repacking

### DIFF
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -19,7 +19,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
     // Read/Write to an .msapp file. 
     internal static class MsAppSerializer
     {
-        internal static readonly string ConnectionInstanceIDPropertyName = "connectionInstanceId";
+        public const string ConnectionInstanceIDPropertyName = "connectionInstanceId";
 
         private static T ToObject<T>(ZipArchiveEntry entry)
         {
@@ -566,8 +566,8 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                             connectionIDReferences.Remove(connection.Key);
                         }
                     }
-                    app._entropy.LocalConnectionIDReferences = null;
                 }
+
                 var json = Utilities.JsonSerialize(app._connections);
                 props.LocalConnectionReferences = json;
             }
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -550,7 +550,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             if (app._connections != null)
             {
                 var connectionIDReferences = app._entropy.LocalConnectionIDReferences;
-                if (connectionIDReferences != null)
+                if (connectionIDReferences != null && connectionIDReferences.Count != 0)
                 {
                     foreach (var connection in app._connections)
                     {
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -12,12 +12,15 @@ using System.IO.Compression;
 using System.Linq;
 using System.Text;
 using System.Text.Json;
+using System.Text.Encodings;
 
 namespace Microsoft.PowerPlatform.Formulas.Tools
 {
     // Read/Write to an .msapp file. 
     internal static class MsAppSerializer
     {
+        internal static readonly string ConnectionInstanceIDPropertyName = "connectionInstanceId";
+
         private static T ToObject<T>(ZipArchiveEntry entry)
         {
             if (entry.Name.EndsWith(".yaml", StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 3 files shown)


src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -527,6 +549,25 @@ private static IEnumerable<FileEntry> GetMsAppFiles(this CanvasDocument app, Err
             var props = app._properties.JsonClone();
             if (app._connections != null)
             {
+                var connectionIDReferences = app._entropy.LocalConnectionIDReferences;
+                if (connectionIDReferences != null)
```
tomjschneider, line 553: I'm guessing this handles the case for backwards compat? Meaning if an older unpacked App does not have this entropy value, this will come back as `null` and be skipped?

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -527,6 +549,25 @@ private static IEnumerable<FileEntry> GetMsAppFiles(this CanvasDocument app, Err
...
+                            JsonElement connectionInstanceID = JsonSerializer.Deserialize<JsonElement>(instanceID);
+
+                            // Deserialized conn instance id is added to the extension data to eventually add it back to properties.json while packing
+                            extensionData.Add(ConnectionInstanceIDPropertyName, connectionInstanceID);
+
+                            // Making sure instance references are removed after adding it to the extension data
+                            connectionIDReferences.Remove(connection.Key);
+                        }
+                    }
+                    app._entropy.LocalConnectionIDReferences = null;
```
tomjschneider, line 569: Just curious, why do we need to `null` this out?

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -295,6 +298,25 @@ public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
...
                     var cxs = Utilities.JsonParse<IDictionary<String, ConnectionJson>>(app._properties.LocalConnectionReferences);
+
+                    foreach (var connectionJson in cxs)
+                    {
+                        var extensionData = connectionJson.Value.ExtensionData;
+                        if (extensionData != null)
+                        {
+                            if (extensionData.TryGetValue(ConnectionInstanceIDPropertyName, out JsonElement connectionInstanceID))
+                            {
+                                var serializedID = JsonSerializer.Serialize(connectionInstanceID);
```
tomjschneider, line 309: Since this is just a string value (as far as I can tell) do we need to JSON serialize/deserialize it?

tomjschneider (commented):
Looks good overall, left a few questions.

src/PAModel/Serializers/MsAppSerializer.cs
NavneetThekkumpat: > Since this is just a string value (as far as I can tell) do we need to JSON serialize/deserialize it?
Good point. So I started without serialization and then found out it throws " invalid start to JSON with character '/' ".Basically since it stores something like a URL, decided to serialize and it works.

src/PAModel/Serializers/MsAppSerializer.cs
NavneetThekkumpat: > I'm guessing this handles the case for backwards compat? Meaning if an older unpacked App does not have this entropy value, this will come back as `null` and be skipped?

I added this as an extra validation.

src/PAModel/Serializers/MsAppSerializer.cs
NavneetThekkumpat: Looks like this line is not needed since key values are removed from connections(line:566) after adding it back to properties. Went ahead and removed it. Thanks for catching it.

src/PAModelTests/DataSourceTests.cs
```diff
@@ -97,7 +97,38 @@ public void TestNoLocalDatabaseRefsWhenLocalDatabaseReferencesPropertyWasEmptyJs
...
+            Assert.IsNotNull(loadedMsApp._entropy.LocalConnectionIDReferences);
+
+            using (var tempFile = new TempFile())
+            {
+                // Repack the app
+                MsAppSerializer.SaveAsMsApp(loadedMsApp, tempFile.FullPath, new ErrorContainer());
+
+                // Comparing original .msapp with repacked .msapp
+                // and testing no deltas 
+                bool ok = MsAppTest.MergeStressTest(pathToMsApp, tempFile.FullPath);
```
tomjschneider, line 127: Talked offline - this is a pretty heavy operation (makes the unit test take 4ish minutes) and I think instead we can evaluate the `ErrorContainer` that comes back from `SaveToSources` on line 113.

src/PAModelTests/DataSourceTests.cs
```diff
@@ -97,7 +97,28 @@ public void TestNoLocalDatabaseRefsWhenLocalDatabaseReferencesPropertyWasEmptyJs
...
+        {
+            var pathToMsApp = Path.Combine(Environment.CurrentDirectory, "Apps", appName);
+            Assert.IsTrue(File.Exists(pathToMsApp));
+            
+            var (msApp, errors) = CanvasDocument.LoadFromMsapp(pathToMsApp);
+            errors.ThrowOnErrors();
+
+            using var sourcesTempDir = new TempDir();
+            var sourcesTempDirPath = sourcesTempDir.Dir;
+            ErrorContainer  errorsCaptured = msApp.SaveToSources(sourcesTempDirPath);
```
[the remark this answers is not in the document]
NavneetThekkumpat: Oops, yes that makes sense. Let me update it again.

src/PAModel/Serializers/MsAppSerializer.cs
tomjschneider: We talked through this - back compat will keep the `connectionInstanceId` as what it was before this change, which is even better.

tomjschneider (approved):

src/PAModelTests/DataSourceTests.cs
```diff
@@ -97,7 +97,28 @@ public void TestNoLocalDatabaseRefsWhenLocalDatabaseReferencesPropertyWasEmptyJs
...
+            errorsCaptured.ThrowOnErrors();
+
+            var loadedMsApp = SourceSerializer.LoadFromSource(sourcesTempDirPath, new ErrorContainer());
```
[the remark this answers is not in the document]
NavneetThekkumpat: Great, refactored the test.

PR #377, opened 2022-03-30
Added Preview messaging for pack & unpack commands

arpavan:
Added a warning message printed in the first line on the console during execution of `--pack` and `--unpack` commands - reminding the user that the tool is still in preview.

### DIFF
--- a/src/PASopa/Program.cs
+++ b/src/PASopa/Program.cs
@@ -17,6 +17,8 @@ namespace PASopa
         {
             Console.WriteLine($"MsApp/Source converter. Version: {SourceSerializer.CurrentSourceVersion}");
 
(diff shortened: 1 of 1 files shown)


src/PASopa/Program.cs
```diff
@@ -17,6 +17,8 @@ static void Main(string[] args)
         {
             Console.WriteLine($"MsApp/Source converter. Version: {SourceSerializer.CurrentSourceVersion}");
 
+            var warningText = "Warning: {0} is in preview, and functionality is not guaranteed. Use extreme caution if using in a production environment. For more information see aka.ms/paccanvas";
```
tomjschneider, line 20: nit: I'd declare this as a static constant outside of `Program`. Just makes it a little easier to find for anyone changing in the future.

tomjschneider (approved):
Looks good - just one nitpick comment, but doesn't block my approval.

PR #381, opened 2022-04-11
Fixing exception from not handling special characters on canvas unpack

landonarogers:
Problem:
An exception is thrown when special characters are present in the filepath, during canvas unpack.

Solution:
When getting the filepath, escape out special characters.

### DIFF
--- a/src/PAModelTests/DataSourceTests.cs
+++ b/src/PAModelTests/DataSourceTests.cs
@@ -120,6 +120,7 @@ namespace PAModelTests
         [DataTestMethod]
         [DataRow(new string[] {"FileNameOne.txt" }, ".txt")]
         [DataRow(new string[] {"FileNameTwo.tx<t" }, ".tx%3ct")]
+        [DataRow(new string[] { "FileNameThr<ee.txt" }, ".txt")]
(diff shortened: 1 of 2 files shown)


src/PAModelTests/DataSourceTests.cs
```diff
@@ -117,6 +117,25 @@ public void TestConnectionInstanceIDHandling(string appName)
...
+            // Make sure exception is not thrown
+            try
+            {
+                FilePath filePath = new FilePath(fileString);
+                filePath.GetExtension();
+            }
+            // If it is thrown, Fail test
+            catch (System.ArgumentException)
+            {
+                Assert.Fail();
```
tomjschneider, line 135: I don't think this will do what you think it will. `Path.GetExtension()` only throws an `ArgumentException` on .NET Framework and .NET Core older than 2.1 ([doc reference](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getextension?view=netcore-3.1)). The unit test project is on .NET Core 3.1, so even if it wasn't encoding the characters it wouldn't throw an exception.

I think you'll need to compare the result from `filePath.GetExtension()` with the espected escaped result.

tomjschneider (approved):
Looks good!

PR #384, opened 2022-04-15
PA3005: "Group control x state refers to non-existent child"

landonarogers:
Top errors are showing multiple results that seem like it is detecting a runtime error. This should not prevent pack\unpack from completing successfully. Should also evaluate other PA3005 to determine if there are other scenarios like this one which we can avoid.

Solution:
Remove exception throw for relevant errors.

### DIFF
--- a/src/PAModel/SourceTransforms/ComponentInstanceTransform.cs
+++ b/src/PAModel/SourceTransforms/ComponentInstanceTransform.cs
@@ -37,12 +37,14 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.SourceTransforms
         private void DoRename(BlockNode control)
         {
             var templateName = control.Name?.Kind?.TypeName ?? string.Empty;
-            if (!ComponentRenames.TryGetValue(templateName, out var rename))
+            if (ComponentRenames.TryGetValue(templateName, out var rename))
+            {
+                control.Name.Kind.TypeName = rename;
+            }
+            else
             {
                 _errors.ValidationError("Renaming component instance but unable to find target name");
-                throw new DocumentException();
             }
-            control.Name.Kind.TypeName = rename;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/PAModel/SourceTransforms/ComponentInstanceTransform.cs
```diff
@@ -37,12 +37,14 @@ public void BeforeWrite(BlockNode control)
...
         {
             var templateName = control.Name?.Kind?.TypeName ?? string.Empty;
-            if (!ComponentRenames.TryGetValue(templateName, out var rename))
+            if (ComponentRenames.TryGetValue(templateName, out var rename))
+            {
+                control.Name.Kind.TypeName = rename;
+            }
+            else
             {
                 _errors.ValidationError("Renaming component instance but unable to find target name");
```
NavneetThekkumpat, line 46: This would still records validation errors(PA3005) I believe. Probably need to handle those too.

landonarogers: We want it to record an error, we just don't want it throwing the exception like it had before (throw new DocumentException();)

NavneetThekkumpat:
These changes look good, are we planning to handle other runtime validation errors too in this PR?

Like I see there are other areas in the repo throwing exception for the same error as src\PAModel\SourceTransforms\ComponentDefinitionTransform.cs: BeforeWrite()

NavneetThekkumpat:
These changes look good, are we planning to handle other runtime validation errors too in this PR?

Like I see there are other areas in the repo throwing exception for the same error as src\PAModel\SourceTransforms\ComponentDefinitionTransform.cs: BeforeWrite()

NavneetThekkumpat:
> These changes look good, are we planning to handle other runtime validation errors too in this PR?
> 
> Like I see there are other areas in the repo throwing exception for the same error as src\PAModel\SourceTransforms\ComponentDefinitionTransform.cs: BeforeWrite()

Actually, on second thought those exceptions might be necessary (template finding/updating in component errors). Approving the changes in this PR so far.

NavneetThekkumpat (approved):

lesaltzm (approved):

PR #387, opened 2022-04-20
Added KnownIssues document

arpavan:
Added Known Issues document, associated with two of the issues recently discovered. The issues listed are final, however the language, descriptions are work in progress. Will be finalized based on feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,8 @@ bld/
 # Uncomment if you have tasks that create the project's static files in wwwroot
 #wwwroot/
 
+.vscode/
+
 # Visual Studio 2017 auto generated files
 Generated\ Files/
 
--- a/.gitignore
+++ b/.gitignore
@@ -37,8 +37,6 @@ bld/
 # Uncomment if you have tasks that create the project's static files in wwwroot
 #wwwroot/
 
-.vscode/
-
 # Visual Studio 2017 auto generated files
 Generated\ Files/
 
@@ -339,6 +337,14 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder
 .mfractor/
+
+# VS Code files for those working on multiple tools
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+*.code-workspace
 
 # Local History for Visual Studio
 .localhistory/
--- a/docs/KnownIssues.md
+++ b/docs/KnownIssues.md
@@ -1,16 +1,16 @@
 # Known Issues
 
-This page lists all the ongoing known issues in PowerApps-Language-Tooling for which we have workarounds. This document will be continuously updated as issues and workarounds are discovered.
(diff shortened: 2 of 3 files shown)


docs/KnownIssues.md
```diff
@@ -0,0 +1,27 @@
+# Known Issues
+
+This page lists all the ongoing known issues in PowerApps-Language-Tooling for which we have workarounds. This document will be continuously updated as issues and workarounds are discovered.
```
tomjschneider, line 3: It could be helpful to have a little more detail here around the root of these issues. These don't necessarily represent "bugs we won't fix" as much as they represent changes that have been made in the underlying .msapp format which can cause roundtrip errors when unpacking or packing an app. These msapps can be brought up to the latest version, which will not encounter a roundtrip error, through the workarounds below.

docs/KnownIssues.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Warning PA2001: Checksum mismatch. AppTests\2.json checksum does not match on extract
+
+This is a known issue with apps developed with a previous version of Studio, and has now been resolved. This error is known to occur due to a stray empty test step that gets added when unpacking the msapp. This can be resolved by adding an action to the empty test step and then deleting the test entirely by the following steps -
+
+1. Open the msapp back into the maker portal
```
tomjschneider, line 9: Not sure if "maker portal" is a term we use in client docs. Looking around it seems like we may want to use the term "Power Apps Studio".

docs/KnownIssues.md
```diff
@@ -0,0 +1,27 @@
...
+4. Now delete the entire test, which should happen successfully
+5. Save the app to local workspace and then unpack.
+
+The app should now unpack successfully without any errors or warnings associated with Tests.
+
+## Error PA3013: Property Value Changed: TopParent.AllowAccessToGlobals (or similar)
+
+This can occur if the AllowAccessToGlobals field (associated with the Access app scope property for a custom component) within the msapp's Components json is not in sync with the same field in the ComponentsMetadata.json file. This can be resolved by the following steps -
+
+1. Open the msapp back into the maker portal
```
tomjschneider, line 21: Similar "maker portal" terminology feedback as previous.

tomjschneider (commented):
Looks great - few minor thoughts. Would also recommend adding something to the README that links to this file.

.gitignore
```diff
@@ -37,6 +37,8 @@ bld/
 # Uncomment if you have tasks that create the project's static files in wwwroot
 #wwwroot/
 
+.vscode/
```
tomjschneider, line 40: Recommend changing this to the standard block from <https://github.com/github/gitignore/blob/main/VisualStudio.gitignore>:

```
# VS Code files for those working on multiple tools
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
*.code-workspace
```

Otherwise we're going to start ignoring the files used for recommended extensions.

tomjschneider (changes requested):
Looks good, just need to update the `gitignore`.

.gitignore
arpavan: @tomjschneider updated gitignore with this suggestion. Thanks for pointing this out!

PR #388, opened 2022-04-21
Empty TestStepsMetadata rule causing hash mismatch

landonarogers:
Problem:
In some cases, when attempting to unpack an msapp using PALT, a TestStepsMetadata rule with empty InvariantScript property is being added after unpack during roundtrip causing a hash mismatch.

Solution:
Don't let a default TestStepsMetadate rule be added.
If one is added, throw an exception.

### DIFF
--- a/src/PAModel/Entropy.cs
+++ b/src/PAModel/Entropy.cs
@@ -85,6 +85,11 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         /// </summary>
         public bool? WasLocalDatabaseReferencesEmpty { get; set; }
 
+        /// <summary>
+        /// Tracks whether TestStepsMetadata is empty or not.
+        /// </summary>
+        public bool? TestStepsMetadataEmpty { get; set; }
+
         // Key is connection id, value is connection instance id
         public Dictionary<string, string> LocalConnectionIDReferences { get; set; } = new Dictionary<string, string>(StringComparer.Ordinal);
 
--- a/src/PAModel/Entropy.cs
+++ b/src/PAModel/Entropy.cs
@@ -88,7 +88,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         /// <summary>
         /// Tracks whether TestStepsMetadata is empty or not.
         /// </summary>
-        public bool? DoesTestStepsMetadataExist = true;
+        public bool? DoesTestStepsMetadataExist { get; set; }
 
         // Key is connection id, value is connection instance id
         public Dictionary<string, string> LocalConnectionIDReferences { get; set; } = new Dictionary<string, string>(StringComparer.Ordinal);
--- a/src/PAModel/Entropy.cs
+++ b/src/PAModel/Entropy.cs
@@ -88,7 +88,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         /// <summary>
         /// Tracks whether TestStepsMetadata is empty or not.
         /// </summary>
-        public bool DoesTestStepsMetadataExist { get; set; }
+        public bool? DoesTestStepsMetadataExist { get; set; }
 
         // Key is connection id, value is connection instance id
         public Dictionary<string, string> LocalConnectionIDReferences { get; set; } = new Dictionary<string, string>(StringComparer.Ordinal);
--- a/src/PAModel/Entropy.cs
+++ b/src/PAModel/Entropy.cs
@@ -88,7 +88,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         /// <summary>
         /// Tracks whether TestStepsMetadata is empty or not.
         /// </summary>
-        public bool? TestStepsMetadataEmpty { get; set; }
+        public bool DoesTestStepsMetadataExist { get; set; }
 
         // Key is connection id, value is connection instance id
         public Dictionary<string, string> LocalConnectionIDReferences { get; set; } = new Dictionary<string, string>(StringComparer.Ordinal);
--- a/src/PAModel/Entropy.cs
+++ b/src/PAModel/Entropy.cs
@@ -88,7 +88,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         /// <summary>
         /// Tracks whether TestStepsMetadata is empty or not.
         /// </summary>
-        public bool? DoesTestStepsMetadataExist { get; set; }
+        public bool? DoesTestStepsMetadataExist = true;
 
         // Key is connection id, value is connection instance id
         public Dictionary<string, string> LocalConnectionIDReferences { get; set; } = new Dictionary<string, string>(StringComparer.Ordinal);
--- a/src/PAModel/SourceTransforms/AppTestTransform.cs
+++ b/src/PAModel/SourceTransforms/AppTestTransform.cs
@@ -139,7 +139,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.SourceTransforms
         public void BeforeWrite(BlockNode control)
         {
             var testStepsMetadata = new List<TestStepsMetadataJson>();
-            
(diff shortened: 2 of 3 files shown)


src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -68,6 +75,9 @@ public void AfterRead(BlockNode control)
 
             foreach (var testStep in testStepsMetadata)
             {
+                // If ever encountering TestStepsMetadata, set empty to false
+                _entropy.TestStepsMetadataEmpty = false;
```
NavneetThekkumpat, line 79: This would be true.

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -45,7 +48,11 @@ public AppTestTransform(CanvasDocument app, ErrorContainer errors, TemplateStore
             _screenIdToScreenName = app._screens
                 .Select(screen => new KeyValuePair<string, string>(app._idRestorer.GetControlId(screen.Key).ToString(), screen.Key)).ToList();
 
+            _entropy = entropy;
             _errors = errors;
+
+            // Say that TestStepsMetadata is empty, by default
+            _entropy.TestStepsMetadataEmpty = true;
```
NavneetThekkumpat, line 55: This is not needed

src/PAModel/Entropy.cs
```diff
@@ -85,6 +85,11 @@ internal class PropertyEntropy
         /// </summary>
         public bool? WasLocalDatabaseReferencesEmpty { get; set; }
 
+        /// <summary>
+        /// Tracks whether TestStepsMetadata is empty or not.
+        /// </summary>
+        public bool? TestStepsMetadataEmpty { get; set; }
```
NavneetThekkumpat, line 91: public bool TestStepsMetadataEmpty

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -130,6 +140,12 @@ public void BeforeWrite(BlockNode control)
         {
             var testStepsMetadata = new List<TestStepsMetadataJson>();
 
+            // If no TestStepsMetadata were added in AfterRead, exit
+            if (_entropy.TestStepsMetadataEmpty == true)
```
NavneetThekkumpat, line 144: if(_entropy.TestStepsMetadataEmpty)

src/PAModel/SourceTransforms/AppTestTransform.cs
landonarogers: This would show that it's not empty, so wouldn't it be false...?

NavneetThekkumpat: oops sorry, so this statement should be inside first if condition and set the value to true

src/PAModel/Entropy.cs
NavneetThekkumpat: Maybe rename it to TestStepsMetadataExists

src/PAModel/SourceTransforms/AppTestTransform.cs
NavneetThekkumpat: or if(!_entropy.TestStepsMetadataExists)

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -130,6 +141,12 @@ public void BeforeWrite(BlockNode control)
         {
             var testStepsMetadata = new List<TestStepsMetadataJson>();
```
NavneetThekkumpat, line 142:
```suggestion
// If TestStepsMetadata does not exist, return early
            if (!_entropy.DoesTestStepsMetadataExist)
            {
                return;
            }
            
            var testStepsMetadata = new List<TestStepsMetadataJson>();
```

src/PAModel/Entropy.cs
```diff
@@ -85,6 +85,11 @@ internal class PropertyEntropy
         /// </summary>
         public bool? WasLocalDatabaseReferencesEmpty { get; set; }
 
+        /// <summary>
+        /// Tracks whether TestStepsMetadata is empty or not.
+        /// </summary>
+        public bool DoesTestStepsMetadataExist { get; set; }
```
tomjschneider, line 91: I think this needs to be a nullable `bool`, similar to `WasLocalDatabaseReferencesEmpty` because it is possible that someone would be packing up a Power App which was previously unpacked before this new entropy value was created. You'll also need to make sure you handle that case in your code that reads in the value.

tomjschneider (dismissed):

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -133,57 +143,84 @@ public void BeforeWrite(BlockNode control)
...
                 if (screenProp != null && !_screenIdToScreenName.ToDictionary(kvp => kvp.Value, kvp => kvp.Key).TryGetValue(screenProp.Expression.Expression, out screenId))
                 {
                     _errors.ValidationError($"Test Step {propName} references screen {screenProp.Expression.Expression} that is not present in the app");
                     throw new DocumentException();
                 }
-
-                testStepsMetadata.Add(new TestStepsMetadataJson()
+                
+                // If TestStepsMetadata exists, add
+                if (_entropy.DoesTestStepsMetadataExist == null)
```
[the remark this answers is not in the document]
NavneetThekkumpat: Good point Tom, but there is a section in AfterRead() above that sets this value to either true or false. So, I believe when this method is invoked DoesTestStepsMetadataExist would either be false or true and cannot be null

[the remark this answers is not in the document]
NavneetThekkumpat: Actually null case needs to be handled (in case of packing)

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -133,57 +143,86 @@ public void BeforeWrite(BlockNode control)
...
```
tomjschneider, line 186: Since you're making this same check in 2 places, just store it in a local variable and use the `??` for null coalescing. So something like 
```cs
bool doesTestStepsMetadataExist = _entropy.DoesTestStepsMetadataExist ?? true;
```

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -133,57 +143,86 @@ public void BeforeWrite(BlockNode control)
...
                 {
-                    Description = Utilities.UnEscapePAString(descriptionProp.Expression.Expression),
-                    Rule = propName,
-                    ScreenId = screenId
-                });
+                    _entropy.DoesTestStepsMetadataExist = true;
+                }
+                
+                if (_entropy.DoesTestStepsMetadataExist == true)
```
tomjschneider, line 191: nit: Don't need to compare a `bool` to `true` (this is also why I'd recommend a local variable with null coalescing.)

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -133,57 +143,86 @@ public void BeforeWrite(BlockNode control)
...
+                }
+            }
+
+            // If TestStepsMetadata exists, add
+            if (_entropy.DoesTestStepsMetadataExist == null)
+            {
+                 _entropy.DoesTestStepsMetadataExist = true;
+            }
+            
+            if (_entropy.DoesTestStepsMetadataExist == true)
```
tomjschneider, line 216: nit: Don't need to compare a bool to true (this is also why I'd recommend a local variable with null coalescing.)

tomjschneider (approved):
Some minor change recommendations but otherwise looks good.

PR #389, opened 2022-04-25
Handling asset path name collision 

NavneetThekkumpat:
Problem: Asset file addition with same path name results in Error PA3001: Internal error. An item with the same key has already been added. ( CanvasDocument: StabilizeAssetFilePaths()).

Solution: Resolving the collision by removing the duplicate key before adding it back.

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -584,7 +584,16 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                 var withoutPrefix = GetAssetFilePathWithoutPrefix(resource.Path);
                 fileEntry.Name = withoutPrefix;
                 _assetFiles.Remove(assetFilePath);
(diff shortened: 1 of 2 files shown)


src/PAModel/CanvasDocument.cs
```diff
@@ -584,7 +584,15 @@ internal void StabilizeAssetFilePaths(ErrorContainer errors)
                 var withoutPrefix = GetAssetFilePathWithoutPrefix(resource.Path);
                 fileEntry.Name = withoutPrefix;
                 _assetFiles.Remove(assetFilePath);
-                _assetFiles.Add(withoutPrefix, fileEntry);
+
```
tomjschneider, line 587: nit: Add a comment or use `dictionary[key] = value` to indicate this is an add/update and we always want the latest value.

tomjschneider (approved):

PR #392, opened 2022-04-27
Handle Null Reference thrown in EnsureFileDirExists

landonarogers:
Handle null reference being thrown in EnsureFileDirExists:

```
System.NullReferenceException: Object reference not set to an instance of an object.

   at Microsoft.PowerPlatform.Formulas.Tools.DirectoryWriter.EnsureFileDirExists(String path)
```

### DIFF
--- a/src/PAModel/Serializers/MsAppSerializer.cs
+++ b/src/PAModel/Serializers/MsAppSerializer.cs
@@ -453,7 +453,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             }
 
             var checksum = new ChecksumMaker();
-            
+
             DirectoryWriter.EnsureFileDirExists(fullpathToMsApp);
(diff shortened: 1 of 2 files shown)


NavneetThekkumpat (approved):

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -435,7 +435,12 @@ public static void SaveAsMsApp(CanvasDocument app, string fullpathToMsApp, Error
         {
             app.ApplyBeforeMsAppWriteTransforms(errors);
 
-            if (!fullpathToMsApp.EndsWith(".msapp", StringComparison.OrdinalIgnoreCase) &&
+            if (String.IsNullOrEmpty(fullpathToMsApp))
```
tomjschneider, line 438: nit: use `string` not `String`

tomjschneider (approved):
Looks good, small nit noted

PR #398, opened 2022-05-02
Updating YamlWriter for non-string types

tomjschneider:
Updating the YamlWriter class to append `=` for serialized values of type `int`, `double`, and `bool`. Without the `=` these will not be valid values in the Power App and have a formula warning when opening the App.

I believe this case was never encountered because `Visit` in PAWriterVisitor.cs turns everything to a string before writing the property:
https://github.com/microsoft/PowerApps-Language-Tooling/blob/e9c7885dbadfb467444e28ab4d15416fa032e28f/src/PAModel/PAConvert/PAWriterVisitor.cs#L109

The only other place these property types are being referenced directly is in YamlPocoSerializer, but this isn't used during the actual serialization of the Power App. I've updated it to be consistent with how it handles `string` types.

All test Apps continue to pass round-trip. Added some unit tests to cover this case as well.

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlWriter.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlWriter.cs
@@ -45,27 +45,39 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
             }
         }
 
-        public void WriteProperty(string propertyName, bool value)
+        public void WriteProperty(string propertyName, bool value, bool includeEquals = true)
         {
             WriteIndent();
             _text.Write(propertyName);
             _text.Write(": ");
+            if (includeEquals)
+            {
+                _text.Write("=");
+            }
             _text.WriteLine(value ? "true" : "false");
         }
 
-        public void WriteProperty(string propertyName, int value)
+        public void WriteProperty(string propertyName, int value, bool includeEquals = true)
         {
             WriteIndent();
             _text.Write(propertyName);
             _text.Write(": ");
+            if (includeEquals)
+            {
+                _text.Write("=");
+            }
             _text.WriteLine(value);
         }
 
-        public void WriteProperty(string propertyName, double value)
+        public void WriteProperty(string propertyName, double value, bool includeEquals = true)
         {
             WriteIndent();
             _text.Write(propertyName);
             _text.Write(": ");
+            if (includeEquals)
+            {
+                _text.Write("=");
+            }
             _text.WriteLine(value);
         }
 
--- a/src/PAModelTests/YamlTest.cs
+++ b/src/PAModelTests/YamlTest.cs
@@ -109,6 +109,48 @@ Obj1:
             Assert.AreEqual(normalizedValue, NormNewlines(p.Value));
         }
 
+
+        [DataTestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public void WriteBool(bool value)
+        {
+            var sw = new StringWriter();
+            var yw = new YamlWriter(sw);
+            yw.WriteProperty("P0", value);
+
+            var t = sw.ToString();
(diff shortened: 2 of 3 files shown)


src/PAModel/PAConvert/Yaml/YamlWriter.cs
```diff
@@ -45,27 +45,39 @@ public void WriteEndObject()
             }
         }
 
-        public void WriteProperty(string propertyName, bool value)
+        public void WriteProperty(string propertyName, bool value, bool includeEquals = true)
```
landonarogers, line 48: Should we make this a single function with a templated type, instead of three identical functions with one different overloaded parameter?

src/PAModelTests/YamlTest.cs
```diff
@@ -109,6 +109,48 @@ public void NewLinesRoundtrip(string value)
...
+        {
+            var sw = new StringWriter();
+            var yw = new YamlWriter(sw);
+            yw.WriteProperty("P0", 1.2);
+
+            var t = sw.ToString();
+            var expected = @"P0: =1.2
+";
+            Assert.AreEqual(expected.Replace("\r\n", "\n"), t.Replace("\r\n", "\n"));
+        }
```
landonarogers, line 152: Could we test all three types in one test with data driven testing?

src/PAModel/PAConvert/Yaml/YamlWriter.cs
landonarogers, line 48: Should we make this a single function with a templated type (for the value parameter), instead of three identical functions with one different overloaded parameter type?

src/PAModelTests/YamlTest.cs
```diff
@@ -109,6 +109,48 @@ public void NewLinesRoundtrip(string value)
...
+            yw.WriteProperty("P0", 12);
+
+            var t = sw.ToString();
+            var expected = @"P0: =12
+";
+            Assert.AreEqual(expected.Replace("\r\n", "\n"), t.Replace("\r\n", "\n"));
+        }
+
+        [TestMethod]
+        public void WriteDouble()
```
landonarogers, line 142: Could we test all three types in one test with data driven testing?

landonarogers (changes requested):

src/PAModel/PAConvert/Yaml/YamlWriter.cs
tomjschneider: We could, and I think it is an interesting idea. That said, I think the intent of this class is to be explicit where possible about what we can serialize correctly. In this case `int`, `double`, `bool`, and `string`. Using a templated type would imply to a consumer that we can handle any type, but that isn't really the case. We have specific logic in the case of `string` and `bool`, and it could be in the future that we have specific logic for `int` and `double` as well, or any new types which are added. For now, I'd prefer to keep these separate and explicit.

src/PAModelTests/YamlTest.cs
tomjschneider: As long as these are separate public methods, I'd prefer they have separate unit tests methods as well.

src/PAModel/PAConvert/Yaml/YamlWriter.cs
akshar24: Do we need separate functions for nullable types? Can the functions defined for non-nullable primitive types handle nullable types? like the one for int can handle int?

tomjschneider: The `int` method cannot handle `int?` types. Within this repo, no, there isn't a current need for nullable types. If we need it for something else we can handle that separately or handle the `null` case in the code calling YamlWriter.

NavneetThekkumpat (approved):

landonarogers (approved):

PR #402, opened 2022-05-04
Handling AllowAccessToGlobals property not in sync

NavneetThekkumpat:
Problem : AllowAccessToGlobals property should be in-sync across custom components, their instances and metadata . Currently there are instance where its not in-sync. Telemetry shows quite a lot of these errors reported.

Solution: Using ComponentDefinitionInfo AllowAccessToGlobals property value over ComponentManifest.AllowAccessToGlobals since ComponentDefinitionInfo values are source of truth

Validation: Roundtrips all test .msapps in test battery and in this repo without errors. Roundtrips validation successful for customer .msapp with this inconsistency.

### DIFF
--- a/src/PAModel/IR/IRStateHelpers.cs
+++ b/src/PAModel/IR/IRStateHelpers.cs
@@ -175,7 +175,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             else
             {
                 templateState = new CombinedTemplateState(control.Template);
-                templateState.ComponentDefinitionInfo = null;
+                templateState.ComponentDefinitionInfo = control.Template.ComponentDefinitionInfo;
                 var templateName = templateState.TemplateDisplayName ?? templateState.Name;
                 templateStore.AddTemplate(templateName, templateState);
             }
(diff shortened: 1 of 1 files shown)


src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -337,15 +337,15 @@ private static (ControlInfoJson.Item item, int index) CombineIRAndState(BlockNod
                     ExtensionData = state.ExtensionData,
                     IsGroupControl = state.IsGroupControl,
                     GroupedControlsKey = state.GroupedControlsKey,
-                    AllowAccessToGlobals = (state.IsComponentDefinition ?? false) ? templateState?.ComponentManifest?.AllowAccessToGlobals : state.AllowAccessToGlobals,
+                    AllowAccessToGlobals = state.AllowAccessToGlobals,
```
Shpakh-MSFT, line 340: >state.AllowAccessToGlobals [](http://example.com/codeflow?start=43&length=26)

Is templateState null for component instances?

Ideally we should always pull this from
templateState?.ComponentManifest?.AllowAccessToGlobals

Is this available in non-compdef scenarios?

jt000: I believe in this case templateState is not null & `templateState.ComponentManifest.AllowAccessToGlobals` doesn't match `state.AllowAccessToGlobals`. For whatever reason, it appears that the `state.AllowAccessToGlobals` is actually the correct value since it is coming directly out of the JSON. If we use the templateState, then it will switch the TopParent.AllowAccessToGlobals to the wrong value.

NavneetThekkumpat: Yup, pulling the property values always from the instances lets us avoid this out of sync situation.

landonarogers (approved):

PR #403, opened 2022-05-05
Handling teststep screen references missing 

NavneetThekkumpat:
Problem: Controls can change screens, screens can be renamed, but teststep value isn't updated and causes errors while roudntripping.

Solution: Convert the errors to warning and using entropy to store the screen id without screens.


src/PAModelTests/ReadTransformTests.cs
```diff
@@ -24,5 +24,22 @@ public void AfterReadNullChildren(string filename)
             msapp.ApplyAfterMsAppLoadTransforms(errorContainer);
             Assert.IsFalse(errorContainer.HasErrors);
         }
+
+        [DataTestMethod]
+        [DataRow("TestStepWithInvalidScreen.msapp")]
+        public void TestStepWithInvalidScreenAfterRead(string filename)
```
jt000, line 30: Could probably make this more re-usable by introducing multiple parameters (filename, has errors, has warnings) which could share the test code with `AfterReadNullChildren` and possibly more...
```
        [DataTestMethod]
        [DataRow("GalleryTemplateNullChildren.msapp", false, false)]
        [DataRow("TestStepWithInvalidScreen.msapp", false, true)]
        public void ApplyAfterMsAppLoadTransforms_Test(string filename, bool hasErrors, bool hasWarnings)
        {
            var path = Path.Combine(Environment.CurrentDirectory, "Apps", filename);
            Assert.IsTrue(File.Exists(path));
            (var msapp, var errorContainer) = CanvasDocument.LoadFromMsapp(path);
            errorContainer.ThrowOnErrors();
            msapp.ApplyAfterMsAppLoadTransforms(errorContainer);
            Assert.AreEqual(errorContainer.HasErrors, hasErrors);
            Assert.AreEqual(errorContainer.HasWarnings, hasWarnings);
        }
```

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -105,16 +105,16 @@ public void AfterRead(BlockNode control)
                 {
                     if (!_screenIdToScreenName.ToDictionary(kvp => kvp.Key, kvp => kvp.Value).TryGetValue(testStep.ScreenId, out var screenName))
                     {
-                        _errors.ValidationError($"ScreenId referenced by TestStep {testStep.Rule} in {control.Name.Identifier} could not be found");
-                        throw new DocumentException();
+                        _errors.ValidationWarning($"ScreenId referenced by TestStep {testStep.Rule} in {control.Name.Identifier} could not be found");
+                        _entropy.RuleScreenIdWithoutScreen[testStep.Rule] = testStep.ScreenId;
```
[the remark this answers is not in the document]
NavneetThekkumpat: That makes sense and yes it looks like it is possible. Thank you for pointing this out. Looking into another possible unique key.

landonarogers (approved):

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -105,16 +105,26 @@ public void AfterRead(BlockNode control)
...
-                        _errors.ValidationError($"ScreenId referenced by TestStep {testStep.Rule} in {control.Name.Identifier} could not be found");
-                        throw new DocumentException();
+                        _errors.ValidationWarning($"ScreenId referenced by TestStep {testStep.Rule} in {control.Name.Identifier} could not be found");
+                        var testStepRuleKey = $"{control.Name.Identifier}.{testStep.Rule}";
+
+                        // checking if this key already exist, not an ideal situation
+                        // Fallback logic just to make sure collisions are avoided.
+                        if (_entropy.RuleScreenIdWithoutScreen.ContainsKey(testStepRuleKey))
+                        {
+                            _errors.ValidationWarning($"RuleScreenIdWithoutScreen has a duplicate key {testStepRuleKey}");
```
jt000, line 115: If we're throwing an exception, should this remain an Error?

src/PAModelTests/ReadTransformTests.cs
```diff
@@ -11,18 +11,21 @@ namespace PAModelTests
 {
     [TestClass]
     public class ReadTransformTests
-    {   
+    {  
         [DataTestMethod]
-        [DataRow("GalleryTemplateNullChildren.msapp")]
-        public void AfterReadNullChildren(string filename)
+        [DataRow("GalleryTemplateNullChildren.msapp", false, false)]
+        [DataRow("TestStepWithInvalidScreen.msapp", false, true)]
```
jt000, line 17: Is this a test for the RuleScreenIdWithoutScreen warning\error?

NavneetThekkumpat: Yes, a test .msapp that reproduced this error and is fixed now.

src/PAModel/SourceTransforms/AppTestTransform.cs
NavneetThekkumpat: Yup makes sense, I can change it to an error and remove the names.

PR #404, opened 2022-05-09
Create Backdoor REPL

jack-work:
Disclaimer:  I realize this update is unprompted and written without team permission, so note that I don't intend to merge it with an sense of expediency.  I am open to any suggestions the team may have.

I needed to procedurally delete screens in order to triage a bug I am investigating.

To do this, I create a basic PASopa REPL that I've called Backdoor (working title).  It unpacks the first argument and starts a REPL for which I implemented a few commands.

`list children control_name` 
Lists all children of `control_name` if it exists

`list screens`
Lists all screens of the application

`delete control_name`
Deletes `control_name` if it exists

`pack location`
Packs current application to provided location

`exit`
Exits the REPL

I originally implemented a menu-based command line, hence the existence of IMenu<T>.  The purpose of this was to provide different functionality depending on user input.

IMenu implements a description, which is emitted to console when the menu is active.  There can only be one active menu at any given time.  A menu implements a transfer function which will evaluate an input and return the next menu.  The transfer function accepts an input and an instance of T as an argument.  It produces a string result and may populate a collection of errors, both of which are displayed to the screen when the input is evaluated.

The ReplMenu is the only menu in this PR.  It will call one of the aforementioned functions depending on the input (e.g. if `list screen` is input, it will invoke the ListControls function.

A function may be called on user input.  A function implements a name so that its implementation may be fetched if the user inputs its name.  A function also implements a method TryDo that bears the implementation of a function.  TryDo accepts a list of arguments that a menu may pass to it on which the function may depend.  It produces a string result and may also populate a collection of errors.

Functions are designed in a telescopic way.  For instance, ListControls bears a list of functions that can be used depending on the second and subsequent arguments.  However, fi they were fed the full list of arguments straight away, they would fail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -357,3 +357,7 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# Jetbrains IDE files
+.idea/
+.run/
--- a/src/Backdoor/Backdoor.csproj
+++ b/src/Backdoor/Backdoor.csproj
@@ -49,7 +49,9 @@
         <Compile Include="Repl\Functions\ListControls.cs" />
         <Compile Include="Repl\Functions\ListFunctions\ListChildren.cs" />
         <Compile Include="Repl\Functions\ListFunctions\ListScreens.cs" />
+        <Compile Include="Repl\Functions\Open.cs" />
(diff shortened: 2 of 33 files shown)


petrochuk:
old PR with conflicts

PR #410, opened 2022-05-11
adding passthrough for custom pages inputs metadata file

milantomic6:
this pr adds ability to copy file containing schma definitions for custom pages that are used by the app - this needs to be passed through for the app that uses Navigate function to navigate to other custom pages.


jt000:
@milantomic6, can you add some details & an MSApp to repro the issue you are fixing? Also, can you add unit tests as a part of this PR?

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -629,6 +639,11 @@ public static void SaveAsSource(CanvasDocument app, string directory2, ErrorCont
                 dir.WriteAllJson("", FileKind.Schema, app._parameterSchema);
             }
 
+            if (app._customPageInputsMetadata != null)
```
NavneetThekkumpat, line 642: @milantomic6: It would be nice to add this piece of code alongside other packagesdir writes. I believe that's around line 545 (after pcf controls) in this method.

NavneetThekkumpat:
Looks good, validated by testing against the battery and test apps. Passes all.

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -627,7 +642,7 @@ public static void SaveAsSource(CanvasDocument app, string directory2, ErrorCont
             if (app._parameterSchema != null)
             {
                 dir.WriteAllJson("", FileKind.Schema, app._parameterSchema);
-            }
+            }
```
lesaltzm, line 645: nit: extra whitespace

src/PAModel/Serializers/SourceSerializer.cs
milantomic6: moved!

src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -251,6 +251,15 @@ public static CanvasDocument LoadFromSource(string directory2, ErrorContainer er
             LoadDataSources(app, dir, errors);
             LoadSourceFiles(app, dir, templateDefaults, errors);
 
+            foreach (var file in dir.EnumerateFiles(PackagesDir))
```
lesaltzm, line 254: Minor: we load other items from packages dir up on line 160, consider specifically looking for this file there, instead of looping all the files in that directory.

src/PAModel/CanvasDocument.cs
```diff
@@ -60,6 +60,7 @@ public class CanvasDocument
         internal ThemesJson _themes;
         internal ResourcesJson _resourcesJson;
         internal ParameterSchema _parameterSchema;
+        internal Dictionary<string, string> _customPageInputsMetadata;
```
lesaltzm, line 63: add a comment about what the content is? i.e.:
// Key is page name, value is ??

lesaltzm (approved):

landonarogers (commented):
Could we add some sort of testing?

landonarogers (approved):

PR #413, opened 2022-05-16
Handling Null ref with ScreenId

NavneetThekkumpat:
Resolving Null ref with ScreenId.
Also improving stresstest to include all test .msapps.

Validation: Roundtrips all test .msapps in test battery and in this repo without errors. 
 Roundtrips validation successful for customer .msapp with this inconsistency.

### DIFF
--- a/src/PAModel/SourceTransforms/AppTestTransform.cs
+++ b/src/PAModel/SourceTransforms/AppTestTransform.cs
@@ -188,7 +188,13 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.SourceTransforms
                 // Lookup screenID by Name
                 if (screenProp != null)
                 {
-                    _screenIdToScreenName.ToDictionary(kvp => kvp.Value, kvp => kvp.Key).TryGetValue(screenProp.Expression.Expression, out screenId);
+                    foreach (var prop in _screenIdToScreenName.ToDictionary(kvp => kvp.Value, kvp => kvp.Key))
+                    {
+                        if (prop.Key != null && prop.Key == screenProp.Expression.Expression)
+                        {
+                            screenId = prop.Value;
+                        }
+                    }
 
                     // in roundtrip scenario screenId could be null
                     if (screenId == null)
--- a/src/PAModel/SourceTransforms/AppTestTransform.cs
+++ b/src/PAModel/SourceTransforms/AppTestTransform.cs
@@ -188,11 +188,11 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.SourceTransforms
                 // Lookup screenID by Name
                 if (screenProp != null)
                 {
-                    foreach (var prop in _screenIdToScreenName.ToDictionary(kvp => kvp.Value, kvp => kvp.Key))
+                    foreach (var prop in _screenIdToScreenName)
                     {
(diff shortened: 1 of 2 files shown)


src/PAModelTests/RoundtripTests.cs
```diff
@@ -13,37 +13,21 @@ namespace PAModelTests
...
-        [DataRow("LocaleBasedApplication.msapp")]
-        public void StressTestApps(string filename)
+        [TestMethod]
+        public void StressTestApps()
         {
-            var root = Path.Combine(Environment.CurrentDirectory, "Apps", filename);
+            var directory = Path.Combine(Environment.CurrentDirectory, "Apps");
+            foreach (var root in Directory.GetFiles(directory))
+            { 
+                Assert.IsTrue(File.Exists(root));
```
jt000, line 22: I'd suggest putting the contents of the foreach in a try\catch where the catch is an `Assert.Fail(ex.ToString())` so you can loop through all even when one of the cases throws

src/PAModel/SourceTransforms/AppTestTransform.cs
```diff
@@ -188,7 +188,13 @@ public void BeforeWrite(BlockNode control)
                 // Lookup screenID by Name
                 if (screenProp != null)
                 {
-                    _screenIdToScreenName.ToDictionary(kvp => kvp.Value, kvp => kvp.Key).TryGetValue(screenProp.Expression.Expression, out screenId);
+                    foreach (var prop in _screenIdToScreenName.ToDictionary(kvp => kvp.Value, kvp => kvp.Key))
```
jt000, line 191: why "toDictionary" if you're just going to look at _screenIdToScreenName's KVPs? Seems like you could leave that off entirely & just swap the Key for Value below...

src/PAModelTests/RoundtripTests.cs
```diff
@@ -13,37 +13,27 @@ namespace PAModelTests
...
-
-            var cloneOk = MsAppTest.TestClone(root);
-            // If this fails, to debug it, rerun and set a breakpoint in DebugChecksum().
-            Assert.IsTrue(cloneOk, $"Clone failed: " + filename);
+                    var cloneOk = MsAppTest.TestClone(root);
+                    // If this fails, to debug it, rerun and set a breakpoint in DebugChecksum().
+                    Assert.IsTrue(cloneOk, $"Clone failed: " + root);
+                }
+            }
+            catch (Exception ex) {
```
jt000, line 34: I'd suggest putting this inside the foreach. This allows for all apps to be tested even when one fails

src/PAModelTests/RoundtripTests.cs
```diff
@@ -13,37 +13,27 @@ namespace PAModelTests
...
+        [TestMethod]
+        public void StressTestApps()
         {
-            var root = Path.Combine(Environment.CurrentDirectory, "Apps", filename);
+            var directory = Path.Combine(Environment.CurrentDirectory, "Apps");
+            try
+            {
+                foreach (var root in Directory.GetFiles(directory))
+                {
+                    Assert.IsTrue(File.Exists(root));
```
jt000, line 24: If Directory.GetFiles returned the file, then I don't think this could ever return `false`

landonarogers (approved):

PR #415, opened 2022-05-17
Handing the same key errors in IRStateHelpers.cs

Super-Sage:
**Problem**:  Dictionary parse same parameter property as key cause "An item with same key has already been added" errors.

**Solution:** Adding control name with the property name as the key to make it unique and resolving the issue.

**Validation**: Roundtrips all test .msapps in test battery and in this repo without errors.


NavneetThekkumpat:
Discussed offline and validated roundtripping test battery, along with testapps in msapps. Looks good

NavneetThekkumpat (approved):

PR #421, opened 2022-06-06
Improving the missing closing quote error log with line details

NavneetThekkumpat:
## Problem
Currently missing closing ' error gives no details on what identifier or cases cause this type of error.

## Solution
Improving the missing closing quote error log with type details to debug the error better
image

## Validation

- Added Unit tests
- Roundtrip testing the test .msapps
- Roundtrip testing the test battery .msapps

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlLexer.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlLexer.cs
@@ -527,19 +527,16 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
             {
                 if (lastSingleQuoteIndex < lineSplit[0].Length)
                 {
-                    if (isComponent)
-                    {
-                         return Error(line, $"Missing closing \' in Component Identifier");
-                    }
-                    return Error(line, $"Missing closing \'in Type node Identifier");
+                    var err = isComponent ? Error(line, $"Missing closing \' in Component Identifier") : Error(line, $"Missing closing \'in Type node Identifier");
+                    return err;
                 }
                 if (isComponent)
                 {
-                    return Error(line, $"Missing closing \' in Component type value");
+                    var err = isComponent ? Error(line, $"Missing closing \' in Component value") : Error(line, $"Missing closing \'in Type node value");
+                    return err;
                 }
-                return Error(line, $"Missing closing \'in Type node type value");
             }
-            return Error(line, $"Missing closing \'in property name");
+            return Error(line, $"Missing closing \'in property");
         }
 
         private YamlToken UnsupportedComment(LineParser line)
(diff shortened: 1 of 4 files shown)


src/PAModel/PAConvert/Yaml/YamlLexer.cs
```diff
@@ -509,6 +515,33 @@ private YamlToken Unsupported(LineParser line, string message)
...
+                    {
+                         return Error(line, $"Missing closing \' in Component Identifier");
+                    }
+                    return Error(line, $"Missing closing \'in Type node Identifier");
+                }
+                if (isComponent)
+                {
+                    return Error(line, $"Missing closing \' in Component type value");
+                }
+                return Error(line, $"Missing closing \'in Type node type value");
```
landonarogers, line 540: What is 'in Type node type value'?

src/PAModel/PAConvert/Yaml/YamlLexer.cs
```diff
@@ -509,6 +515,25 @@ private YamlToken Unsupported(LineParser line, string message)
             return Error(line, message);
         }
 
+        private YamlToken UnsupportedSingleQuote(LineParser line, int lastSingleQuoteIndex, bool isComponent)
+        {
+            string[] lineSplit = line._line.ToString().Split(new string[] { "As" }, StringSplitOptions.None);
```
tomjschneider, line 520: Small thing, but I'd recommend doing a `ToLower()` here and comparing against "as" just in case the casing on this changes at some point.

tomjschneider, line 520: You'll also need to split on " as " here, I think. Otherwise you could catch the characters "as" is they are part of a control name or the name of a control definition. Think about the case `UnpastedControl as screen` - you'd end up with "Unp", "tedControl ", and " screen" as your splits.

NavneetThekkumpat: Good catch. I shall make the changes

src/PAModel/PAConvert/Yaml/YamlLexer.cs
```diff
@@ -509,6 +518,21 @@ private YamlToken Unsupported(LineParser line, string message)
             return Error(line, message);
         }
 
+        private YamlToken UnsupportedSingleQuote(LineParser line, bool isComponent)
+        {
+            string[] lineSplit = line._line.ToString().ToLower().Split(new string[] { " as " }, StringSplitOptions.None);
```
tomjschneider, line 523: Why split on a `string[]` with one element instead of just splitting on the string itself?

tomjschneider (approved):
Some small suggestions but looks good otherwise.

src/PAModel/PAConvert/Yaml/YamlLexer.cs
NavneetThekkumpat: It seems like split take a char, char array or string array as arguments. Unless I am missing something.

tomjschneider: Oh yeah - must be a .NET Standard thing that it doesn't support the string overload. Works for me!

PR #423, opened 2022-06-13
Modify checkDuplicate error message

landonarogers:
Created to modify the error message being printed by checkDuplicate.

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlLexer.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlLexer.cs
@@ -726,7 +726,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
                 if (_previousProperties.TryGetValue(propName, out oldLine))
                 {
                     // Key is already present.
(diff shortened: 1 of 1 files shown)


src/PAModel/PAConvert/Yaml/YamlLexer.cs
```diff
@@ -726,7 +726,7 @@ internal YamlToken CheckDuplicate(string propName, int currentLine)
                 if (_previousProperties.TryGetValue(propName, out oldLine))
                 {
                     // Key is already present.
-                    return YamlToken.NewError(default(SourceLocation), $"Property '{propName}' is already defined on line {oldLine}.");
```
jt000, line 729: Can you include both the old & new line? Maybe something like "Property name duplicated on lines {oldLine} and {newLine}"?

landonarogers: Done! :)

Super-Sage (approved):

PR #424, opened 2022-06-27
Handling Newtonsoft Component Governance

NavneetThekkumpat:
## Problem

PALT - Component Governance - Newtonsoft 9.0.1

## Solution

Update Newtonsoft 9.0.1 to 13.0.1

## Validation
Ran the build locally (version updated locally)
image

### DIFF
--- a/src/PAModelTests/PAModelTests.csproj
+++ b/src/PAModelTests/PAModelTests.csproj
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
     <PackageReference Include="MSTest.TestAdapter" Version="2.1.0" />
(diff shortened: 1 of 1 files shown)


src/PAModelTests/PAModelTests.csproj
```diff
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
```
arpavan, line 13: There is a chance that when the PR action runs, it may still go down the dependency chain for the `Microsoft.NET.Test.Sdk` version `16.5.0` and throw a warning for this issue. Let's go ahead with this change (possibly try consuming into one of the products on a branch, where the test is triggered) and keep an eye out for this warning.

arpavan (approved):

arpavan:
Approved with comments.

src/PAModelTests/PAModelTests.csproj
NavneetThekkumpat: Possible, good point.

PR #427, opened 2022-06-29
handling null ref exception in RestoreAssetFilePaths

akshar2401:

## Problem

Null Reference Exception in CanvasDocument.RestoreAssetFilePaths

## Solution

Adding null checks and taking appropriate actions

## Changes
-Added null check conditions in CanvasDocument.RestoreAssetFilePaths

## Validation

Unit testing

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -2,19 +2,17 @@
 // Licensed under the MIT License.
 
 using Microsoft.AppMagic.Authoring.Persistence;
-using Microsoft.PowerPlatform.Formulas.Tools.IR;
+using Microsoft.PowerPlatform.Formulas.Tools.ControlTemplates;
 using Microsoft.PowerPlatform.Formulas.Tools.EditorState;
+using Microsoft.PowerPlatform.Formulas.Tools.IR;
+using Microsoft.PowerPlatform.Formulas.Tools.Schemas;
+using Microsoft.PowerPlatform.Formulas.Tools.Schemas.PcfControl;
+using Microsoft.PowerPlatform.Formulas.Tools.SourceTransforms;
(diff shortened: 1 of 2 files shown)


src/PAModel/CanvasDocument.cs
```diff
@@ -625,16 +623,19 @@ private int FindMaxEntropyFileName()
         private void RestoreAssetFilePaths()
         {
             // For apps unpacked before this asset rewrite was added, skip the restore step
-            if (_entropy.LocalResourceFileNames == null)
+            if (_entropy?.LocalResourceFileNames == null)
```
NavneetThekkumpat, line 626: You would still need to separate it out `if (_entropy == null || _entropy.LocalResourceFileNames == null)`

src/PAModel/CanvasDocument.cs
```diff
@@ -625,16 +623,19 @@ private int FindMaxEntropyFileName()
...
                 return;
 
-            if (_resourcesJson == null || _assetFiles == null)
+            if (_resourcesJson?.Resources == null || _assetFiles == null)
```
NavneetThekkumpat, line 629: You would still need to separate it out `if (_resourcesJson== null || _resourcesJson.Resources == null || _assetFiles == null)`

src/PAModel/CanvasDocument.cs
akshar2401: I think null coalescing handles both scenarios. I think _entropy?.LocalResourceFileNames will be resolved to null by   short circuiting to null when _entropy itself is null or yielding null as the final result if _entropy is not null, but LocalResourceFileNames is null. If the preferred style is to break it down into two separate comparisons, then I can break the expression down and update the pr accordingly.

NavneetThekkumpat (approved):

PR #430, opened 2022-07-07
Resolve SourceSerializer Sticking in Loop

landonarogers:
## Problem
Resolving potential issue that arises when filenames could collide. The logic intended to resolve this can potentially get stuck in a loop.

## Solution

Add line to allow while-loop to break at appropriate time.

## Validation

- Ran against code that previously got loop. No longer gets stuck in loop.


paulbreuler (commented):
Thanks for the fix!

FYI: To apply this to PAC we will need to bump the version from 0.5.1-preview to the version with this fix.

tomjschneider:
This looks like a good potential use case for a unit test, if it is feasible to add one.

NavneetThekkumpat (approved):
Looks good. Agree on unit test

Issue #428, opened 2022-06-30
PA3013: Property Value Changed: TopParent.Children[0].Children[0].Rules[2].InvariantScript


neversleepz:
- [ ] Please attach a sample .msapp to this Issue which demonstrates the issue you are experiencing. *(Please do not attach any file which may contain sensitive personal or company data.)*

I can email the file or provide a link through our companies fileshare service (will need an email address)

**Describe the bug**

Using the pac cli, I'm getting PA3013 error and a PA2001 error when trying to unpack an msapp I'm working on

**To Reproduce**

See below

**Screenshots**

```
PS Y:\Downloads> pac canvas unpack --msapp '.\Member Details - Development Copy.msapp' --sources MemberDetailsDC
Warning: canvas unpack is in preview, and functionality is not guaranteed. Use extreme caution if using in a production environment. For more information see https://aka.ms/paccanvas
Unpack from Y:\Downloads\Member Details - Development Copy.msapp --> Y:\Downloads\MemberDetailsDC
  source format version: 0.24
Warning PA2001: Checksum mismatch. Checksum indicates that sources have been edited since they were unpacked. If this was intentional, ignore this warning.
Microsoft PowerPlatform CLI
Version: 1.16.6+g21922f9

Error: Error   PA3013: Property Value Changed: TopParent.Children[0].Children[0].Rules[2].InvariantScript
Error   PA3011: Roundtrip validation on unpack failed.
You have found a bug; this is not a specific bug, rather an indicator that some bug has been encountered.
Please open an issue and log the entirety of this error log at https://github.com/microsoft/PowerApps-Language-Tooling

```

**Additional context**

Started off as a simple PowerApp created from Excel (Portrait Phone Layout).  I did however have to manually edit the msapp previously to convince the app it should really be a Desktop laid out screens per this blog post: https://powerusers.microsoft.com/t5/forums/forumtopicprintpage/board-id/PowerAppsForum1/message-id/127741/print-single-message/true/page/1
(This may be the cause of the mismatch warning)

The app itself is quite simple, intended so that non-technical users can edit the excel datasources and add fields to the existing forms in the powerapp using the Form Control Edit Fields.  It does however have a lot of form fields.  
It also does call execute an external Power Automate Flow or two.  Uses a Sharepoint Directory Connector too, and a PDF Reader component.

tomjschneider:
[...]

PR #435, opened 2022-08-11
Fixing serialization inconsistency

tomjschneider:
Related to [Issue 428](https://github.com/microsoft/PowerApps-Language-Tooling/issues/428).

## Problem

Roundtrips are failing during an unpack or pack when the description of a test step contains certain Unicode characters (such as `U+00A0`, a non-breaking space.)

This is caused by a difference between how an .msapp saved from Canvas is serialized and how an .msapp which is packed from Formulas.Tools is serialized. 

## Solution

During its initial serialization, I am changing the serialization of `TestStepsMetadata` to use `Newtonsoft.Json`, which matches the logic in Canvas serialization. Our level of control around character encoding when using `System.Text.Json` is limited by its [global block list](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-character-encoding#global-block-list).

I considered two other alternatives:

1. Manually replace the escaped Unicode after the initial serialization of `TestStepsMetadata`. This fixes the particular issue we are seeing, but has the potential to miss other use cases which we haven't yet encountered since it still is not using exactly the same serialization as Canvas. See [this PR](https://github.com/microsoft/PowerApps-Language-Tooling/pull/434).
2. Implement a custom `JavaScriptEncoder` and use that in the Json serialization. The two issues I had with this were that it is a pretty heavy-handed approach with potential performance implications, and it would involve needing to compile code with the `unsafe` keywork. See [this issue](https://github.com/dotnet/runtime/issues/54193) for details on implementing a custom `JavaScriptEncoder`.

I am open to feedback on alternate or more robust approaches.

## Changes

- Updated `AppTestTransform` to serialize and deserialize `TestStepsMetadata` using `Newtonsoft.Json`.

## Validation

I was able to reproduce the issue and added an .msapp to our suite of app tests.


src/PAModelTests/PAModelTests.csproj
```diff
@@ -41,6 +41,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="Apps\UnicodeEscapingInTestSteps.msapp" />
```
NavneetThekkumpat, line 44: Guess this was auto added. Might be good to remove this line

tomjschneider: You're right. I almost snuck it in before you found it 😄 Fixed now.

NavneetThekkumpat (approved):

PR #437, opened 2022-08-23
Reduced Error Message Body Length Avoiding Truncated Error Telemetry

arpavan:
## Problem
The `message` column in our internal telemetry databases has limited maximum length, and due to this, certain errors that were logged by this application were getting truncated because the message body was too long (by several hundreds of orders of magnitude) and therefore this was resulting in bad telemetry data for errors.

This was happening because the application was appending every single error message during a command execution to the error message body before publishing the error object. In some cases, like canvas `pack` and `unpack` where the round trip errors could lead to 100s of lines of errors, this was causing the entire message body to get truncated, leading to loss of valuable telemetry data used in subsequent columns.

## Solution
To fix this, we have borrowed inspiration from the .NET exception message, and have implemented a similar solution here - Now the error output restricts itself to 10 error lines, and for all the excess lines, only a count is reported E.g. "...and 28 more errors". This ensures that no matter how many errors a particular action result has, we always have a fixed length error message body, that also informs the user that there are more errors number that need to be resolved after the first 10.

image

As a side effect, this change applies the same solution to the error logs that go to the console as well, this _could_ be preferred since it maintains consistency across all of the error log outputs, and is similar to behavior seen elsewhere in error log outputs across .NET platforms.

## Changes
- The error container function `AddError` now also checks if the `_errors` collection is bigger than 10 objects, and if so skips adding any more error objects to the collection. It instead just keeps track of the count of errors that it attempts to add
- The `AddError` function also now keeps track of the total number of errors and warnings and saves them to newly introduced private variables `_countErrors` and `_countWarnings` - these were local variables in the `Write` function previously.
- The `Write` function now outputs a maximum of 10 error lines and once this limit has reached, it will stop and print a statement "...and X more errors" - where X is the count of the additional errors in the collection.
- The limit of 10 is set to a constant private variable `_errorsToPrint` - the number 10 has been arbitrarily determined to keep the output size to a suitable level.

## Validation
Unit test `TestMessageWithTooManyErrors` has been added to check for the error output. Input data for less, limit, and excess error lines with the expected output has been added to ensure this works. However, we would need to run this against a Kusto environment to fully validate this change.

### DIFF
--- a/src/PAModel/PAConvert/ErrorContainer.cs
+++ b/src/PAModel/PAConvert/ErrorContainer.cs
@@ -1,10 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-// TODO: Sort Imports
 using Microsoft.PowerPlatform.Formulas.Tools.IR;
-using Microsoft.PowerPlatform.Formulas.Tools.Parser;
-using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
@@ -27,7 +24,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         internal void AddError(ErrorCode code, SourceLocation span, string errorMessage)
         {
             var error = new Error(code, span, errorMessage);
-            // TODO: Capture count in a new private variable
+
             if (error.IsError) { _countErrors++; } else { _countWarnings++; }
 
             if (_errors.Count < _errorsToPrint)
@@ -66,14 +63,13 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
         {
             foreach (var error in this)
             {
-                // if (error.IsError) { countErrors++; } else { countWarnings++; }
                 output.WriteLine(error);
             }
 
             if (_countErrors > _errorsToPrint)
             {
                 var additionalErrors = _countErrors - _errorsToPrint;
-                output.WriteLine($"...and {additionalErrors} errors.");
+                output.WriteLine($"...and {additionalErrors} more errors.");
             }
 
             if (_countErrors + _countWarnings > 0)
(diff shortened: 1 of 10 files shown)


src/PAModel/PAConvert/ErrorContainer.cs
```diff
@@ -51,18 +60,20 @@ IEnumerator IEnumerable.GetEnumerator()
         // Helper for writing out errors.
         public void Write(TextWriter output)
```
jt000, line 61: Is the problem in the console output? The bug is related to the telemetry in PA-Client, which I had thought was logged from a different location (not from console). I don't think there's a need to limit the console output that I know of...

arpavan: This change has been made within `ErrorContainer` itself which is then passed onto wherever the errors need to be logged. It can be console, or in this case even Kusto (if I'm not mistaken) i.e., the error container (the source) is still common irrespective of where we're sending these logs.

tomjschneider:
This change appears to be modifying the functionality of this library to compensate for a limitation of a consumer. This impacts the error output of anyone using this library, which includes our own team when trying to troubleshoot issues. Without modifying the code, we'll no longer be able to see the full error output when an issue occurs with an App, if the number of errors is high enough.

I'm not opposed to the idea of a design change around how we report out errors to make them more consistent with other Microsoft libraries, and it isn't always valuable to get back 100s of errors in output. That said, I think that type of design change needs more consideration and discussion. For example, right now you have `_errorsToPrint` controlling the number of errors we will output when `ErrorContainer` writes its error messages and also controlling the number of errors we will log in the `ErrorContainer` overall. I'm not sure this second usage, and limiting the errors we collect in the container, is something we want to do. We should also consider making this value configurable, rather than hardcoded in as it is now.

As a stopgap solution, could we not update PA-Client's telemetry code to cut the message off at whatever limit it has in terms of the string length? This feels to me like an issue that should be resolved on the consumer side.

arpavan:
Per today's review, this change will be abandoned here and implemented on the consumer side.

PR #438, opened 2022-08-25
adding checks for possible nulls and logging any unhandled null ref a…

akshar2401:
…s internal error

## Problem

Null Ref Exception was logged in the dashboard in the MsAppSerializer.GetMsAppFiles() function. The function is huge so it is extremely difficult to figure out what cause null ref exception. Tried reproducing it but no success.

## Solution

Added null checks for possible cases. Null ref was generated in different functions because of these cases in the past. Wrapped the code in try/catch to catch the null ref and log it as an internal error which will reveal more like details like line number on which null ref occured

## Changes
Added null checks for possible cases of null ref
Wrapped code in try catch to log null ref as internal error which will reveal more details to debug if null ref occurs again

## Validation

Unit Tests

### DIFF
--- a/src/PAModelTests/WriteTransformTests.cs
+++ b/src/PAModelTests/WriteTransformTests.cs
@@ -81,5 +81,33 @@ namespace PAModelTests
             Assert.IsTrue(errors.HasErrors);
             Assert.IsNotNull(errors.Where(error => error.Code == ErrorCode.InternalError).FirstOrDefault());
         }
+
+        [DataTestMethod]
+        [DataRow("AccountPlanReviewerMaster.msapp")]
+        public void TestNullExceptionInGetMsAppFilesIsLoggedAsAnInternalError(string filename)
+        {
+            var root = Path.Combine(Environment.CurrentDirectory, "Apps", filename);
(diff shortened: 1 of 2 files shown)


src/PAModelTests/WriteTransformTests.cs
```diff
@@ -81,5 +81,33 @@ public void TestNullExceptionInRestoreAssetsFilePathsIsLoggedAsAnInternalError(s
...
+        [DataTestMethod]
+        [DataRow("AccountPlanReviewerMaster.msapp")]
+        public void TestNullExceptionInGetMsAppFilesIsLoggedAsAnInternalError(string filename)
+        {
+            var root = Path.Combine(Environment.CurrentDirectory, "Apps", filename);
+            Assert.IsTrue(File.Exists(root));
+
+            (var msapp, var errors) = CanvasDocument.LoadFromMsapp(root);
+            errors.ThrowOnErrors();
+
```
NavneetThekkumpat, line 94: nit: extra line

akshar2401: Done

PR #439, opened 2022-09-15
Directly pass through custom page schema files, and handle any validation in canvas

lesaltzm:
## Problem

Related to https://github.com/microsoft/Power-Fx/pull/663, we're consolidating the type definition logic for this, and using yaml as the native format for it in the .msapp. As such, the .PALT tool does not need to do anything special to handle this file. 

## Solution
Removes special handling for schema.yaml

## Changes
Removes special handling for schema.yaml, associated POCOs, and passes the value through as text. 

## Validation
Testing integration with related changes is ongoing and this will not be checked in until all are ready.


tomjschneider:
I'll keep an eye on this. One thing to consider is ensuring that existing unpacked Apps can still be packed successfully after this change is in place, and if we need any special handling for that.

lesaltzm:
> I'll keep an eye on this. One thing to consider is ensuring that existing unpacked Apps can still be packed successfully after this change is in place, and if we need any special handling for that.

This particular change only applies to an unshipped feature that was never exposed to customers outside of the Test environment, so we shouldn't need to worry about back compat here.

tomjschneider (approved):

PR #446, opened 2022-10-18
Handling localdatabase references when unpacking

NavneetThekkumpat:
## Problem
Error on unpack with unused Dataverse data source

## Problem Area
-  Data source references with empty data sources were skipped.
- In some cases, data source can be empty, but the extension data is important and lost while unpacking causing the roundtrip to fail.
image
 
## Solution
- Make sure LocalDatabaseRef contents are not lost while unpacking.
- Just save the LocalDatabaseRef as it is without skipping any content.

## Validation
- Roundtrip validation on user app
- Unit testing
- Stress test on all test apps successful

### DIFF
--- a/src/PAModel/Serializers/SourceSerializer.cs
+++ b/src/PAModel/Serializers/SourceSerializer.cs
@@ -822,7 +822,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             app._dataSourceReferences = new Dictionary<string, LocalDatabaseReferenceJson>();
             foreach (var dataSourceReferenceEntry in dataSourceReferences)
             {
-                if (dataSourceReferenceEntry.Value.dataSources.Count > 0)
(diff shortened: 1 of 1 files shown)


tomjschneider:
Looks like @akshar2401 added this code a while back to handle a specific scenario. He did add an app to test it, so as long as that is passing, I assume we should be fine. But worth a check with him to make sure this isn't reintroducing something we're not thinking about.

akshar2401:
This looks fine. Can you validate it with the app for which this fix was added?

akshar2401:
> > This looks fine. Can you validate it with the app for which this fix was added?
> 
> Yes, that was validated. All test apps roundtrips well too.

I added a property in entropy to check if localdbreferences was "" or {}. this property would not be present in older apps so this check that you fixed was added to address backward compat. I am worried that apps that are way older might run into the similar issue that this check fixed

NavneetThekkumpat:
Discussed it offline and validated all changes. All good!

PR #451, opened 2022-12-02
Handling Control instances with same control template but different field values

NavneetThekkumpat:
## Problem
Handling Control instances with same control template but different  OverridableProperties and PCFDynamicSchemaForIRRetrieval values.

## Solution
Storing Template PCFDynamicSchemaForIRRetrieval/OverridableProperties field values for each control instance.
Since this value could be different for each control instance though it follows same control template.
 Eg:Control Instance 1 -> template1 -> PCFDynamicSchemaForIRRetrieval1/OverridableProperties1
    Control Instance 2 -> template1 -> PCFDynamicSchemaForIRRetrieval2/OverridableProperties2

Test app with these properties added.

## Validation
- Unit testing
- Roundtrip test


src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -179,6 +179,19 @@ private static void SplitIRAndState(ControlInfoJson.Item control, string topPare
...
+            // Storing Template PCFDynamicSchemaForIRRetrieval/OverridableProperties field values for each control instance.
+            // Since this value could be different for each control instance though it follows same control template.
+            // Eg:Control Instance 1 -> template1 -> PCFDynamicSchemaForIRRetrieval1/OverridableProperties1
+            // Control Instance 2 -> template1 -> PCFDynamicSchemaForIRRetrieval2/OverridableProperties2
+
+            if (control.Template.ExtensionData.TryGetValue("PCFDynamicSchemaForIRRetrieval", out object val))
+            {
+                entropy.PCFDynamicSchemaForIRRetrievalEntry.Add(control.Name, val);
+            }
+            if (control.Template.ExtensionData.TryGetValue("OverridableProperties", out object val1))
```
landonarogers, line 191: nit: Do we want to be more specific with variables names instead of adding numbers to them? Maybe changing the name of 'val' and 'val1' to something like 'PDFVal' and 'propertyVal'?

src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -423,9 +436,20 @@ private static (ControlInfoJson.Item item, int index) CombineIRAndState(BlockNod
                 resultControlInfo.HasDynamicProperties = hasDynamicProperties;
                 resultControlInfo.AllowAccessToGlobals = templateState?.ComponentManifest?.AllowAccessToGlobals;
             }
-            resultControlInfo.Template = template;
+            resultControlInfo.Template = template.JsonClone();
             resultControlInfo.Children = orderedChildren;
 
+            // Using the stored PCFDynamicSchemaForIRRetrieval/OverridableProperties value for each control instance,
+            // instead of the default value from the control template.
+            if (entropy.OverridablePropertiesEntry.TryGetValue(controlName, out object val))
```
landonarogers, line 444: nit: Could creating an out object of 'val' here be an issue, or at least redundantt/less-readable? An out object name 'val' already exists above, so we're overloading this variable

src/PAModel/IR/IRStateHelpers.cs
NavneetThekkumpat: Good catch! makes sense, need to clean up the names. Forgot to change.

landonarogers (approved):

Issue #449, opened 2022-10-25
"PA3001: Internal error. '0x89' is an invalid start of a value" when unpacking msapp


tomjschneider:
Running unpack on the provided msapp results in the following error:
```
Error   PA3001: Internal error. '0x89' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
Stack Trace:
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
   at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
   at System.Text.Json.Utf8JsonReader.Read()
   at System.Text.Json.JsonDocument.Parse(ReadOnlySpan`1 utf8JsonSpan, Utf8JsonReader reader, MetadataDb& database, StackRowStack& stack)
   at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 utf8Json, JsonReaderOptions readerOptions, Byte[] extraRentedBytes)
   at System.Text.Json.JsonDocument.Parse(ReadOnlyMemory`1 utf8Json, JsonDocumentOptions options)
   at Microsoft.PowerPlatform.Formulas.Tools.MsAppTest.FlattenJson(Byte[] json) in C:\Repos\PowerApps-Language-Tooling\src\PAModel\MsAppTest.cs:line 308
   at Microsoft.PowerPlatform.Formulas.Tools.MsAppTest.CompareChecksums(String pathToZip, TextWriter log, Dictionary`2 comp, Boolean first, ErrorContainer errorContainer) in C:\Repos\PowerApps-Language-Tooling\src\PAModel\MsAppTest.cs:line 280
   at Microsoft.PowerPlatform.Formulas.Tools.MsAppTest.Compare(String pathToZip1, String pathToZip2, TextWriter log, ErrorContainer errorContainer) in C:\Repos\PowerApps-Language-Tooling\src\PAModel\MsAppTest.cs:line 235
[...]

Issue #440, opened 2022-09-16
NullRefException when unpacking .msapp


tomjschneider:
**Describe the bug**

See https://github.com/microsoft/powerplatform-build-tools/issues/194

Unpacking customer app throws a roundtrip error and a NullReferenceException. Sample app provided to team - can be reproduced in latest PASopa.

**To Reproduce**

1. Unpack provided msapp
2. Observe NullReferenceException

PR #455, opened 2022-12-14
Updating asset collision handling

tomjschneider:
## Problem

The current logic for unpacking and packing an App can run into issues if there are collisions between the existing and the renamed files. The code right now tried to manage these in place, but there are scenarios where the new or old name of the file causes the previous entry to be overwritten and lost. An example of this can be found in #449.

The existing workflow roughly looks like this:
1. When we convert the app from an MSApp to source files (unpacking), we take all the images and rename them. Instead of the names they were given in the MSApp, they get named to be the friendly name of the image. Take my example app, these become:
   1. 0002.png -> Image2.png
   1. 8a8c1695-20fc-426f-a4c1-95984c977124.png -> 0001.png
   1. 548996b7-0cd8-4a9a-a417-351c30cc2da6.png -> 0002.png
1. Now in our unpacked app we have all the images as their friendly names.
1. When we go to pack the app back up, we take all these files and put them into `_assetFiles` in the CanvasDocument before eventually calling `RestoreAssetFilePaths()`.
1. Inside this function, it goes through each image and it:
   1. Finds the previous name of the image (so the reversal of my 3 files above.)
   1. Removes the friendly name entry from _assetFiles
   1. Adds or updates `_assetFiles` with the original filename

In Step 4, imagine the following:
1. `_assetFiles` contains our 3 unpacked images: Image2.png, 0001.png, 0002.png
1. We find a file called Image2.png, whose original name was 0002.png
1. We remove Image2.png from the `_assetFiles` and add 0002.png. Problem: we just overwrote the existing 0002.png.
1. Now we find a file called 0002.png, whose original name was 548996b7-0cd8-4a9a-a417-351c30cc2da6.png.
1. We remove 0002.png (problem: we just added this file) and add 548996b7-0cd8-4a9a-a417-351c30cc2da6.png. We just lost a file.

## Solution

I am trying to simplify this and resolve collisions by keeping track of the new asset file entries separately, and then resolving them once we are finished processing the resources, instead of doing the swaps in place. This is a little less efficient, but I believe it will prevent any potential collision issues.

While fixing this, I refactored the unpacking portion as well inside `StabilizeAssetFilePaths()`, since it was causing my sample App to fail a Stress Test. It was similarly having issues where paths were being overwritten in `_assetFiles`, it just appears that those issues were never presenting as a roundtrip error in our existing test Apps.

## Validation

All unit tests continue to pass, and I have added a new roundtrip App to test this specific scenario that identified this issue.


NavneetThekkumpat:
Looks great, confirms this fix works well also by testing the scenario reported in an earlier issue #440.
That issue involved asset collision and involved similar area of code.

NavneetThekkumpat (approved):
Looks great, confirms this fix works well also by testing the scenario reported in an earlier issue https://github.com/microsoft/PowerApps-Language-Tooling/issues/440.
That issue involved asset collision and similar area of code.

PR #456, opened 2022-12-17
Check if JSON before Parsing in mismatch check

landonarogers:
Issues are caused because PNG files are being parsed as JSON objects, when checking if files are mismatched.

## Solution

Do not parse as JSON if not valid JSON.

## Changes

- Parsed in a try catch block to validate if valid JSON

## Validation

- Create Unit Test


src/PAModel/MsAppTest.cs
```diff
@@ -276,18 +277,35 @@ public static void CompareChecksums(string pathToZip, TextWriter log, Dictionary
...
+                                        JsonDocument.Parse(newContents);
+                                    }
+                                    catch
+                                    {
+                                        mismatch = true;
+                                    }
+
+                                    if (!mismatch)
+                                    {
+                                        var jsonDictionary1 = FlattenJson(originalContents);
```
NavneetThekkumpat, line 297: We could ideally move all the contents in this if block within the same try block above, right?
If JsonDocument.Parse throws exception, your code would catch it anyways I believe.

landonarogers: The reason I made the try catch separate, is that, we want to make sure the exception being caught has to do with the JSON being invalid. Adding the lines below to the try block will make it so that 'any' exception will be caught, and not continue. We still want those to be handled properly

src/PAModelTests/ChecksumTests.cs
```diff
@@ -7,13 +7,29 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+using System.Reflection.PortableExecutable;
 using System.Text;
 
+
```
NavneetThekkumpat, line 13: nit: extra line

NavneetThekkumpat (approved):

tomjschneider (approved):
nit: Would recommend putting unit test in `ErrorTests.cs` since it is an error checking test, not a checksum test. Looks good though.

PR #461, opened 2022-12-19
Resolving multiple NullRef errors

landonarogers:
## Problem

Resolving several 'Object Reference not set to an instance of an object' issues:
- _templates in AppSerializer was not being checked if null
- _publishInfo in TransformLogo was not being checked if null
- _properties in AppSerializer was not being checked if null
- _resourcesJson in CanvasDocument was not being checked if null
- ThemesJson's Assert check was not actually preventing ThemesJson from being null
- documentProperties passed in func in Entropy was not being checked for null

## Solution
- Make sure code checks for items that could be null, before querying them.
-  Should all of these be allowed as null? If not, we can add additional Integrity Checks section of OnLoadComplete() in Canvas Document

## Validation

- Add unit testing

### DIFF
--- a/src/PAModel/CanvasDocument.cs
+++ b/src/PAModel/CanvasDocument.cs
@@ -351,17 +351,23 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
 
             // Shard templates, parse for default values
             var templateDefaults = new Dictionary<string, ControlTemplate>();
-            foreach (var template in _templates.UsedTemplates)
+
+            if (_templates != null)
             {
-                if (!ControlTemplateParser.TryParseTemplate(_templateStore, template.Template, _properties.DocumentAppType, templateDefaults, out _, out _))
+                foreach (var template in _templates.UsedTemplates)
                 {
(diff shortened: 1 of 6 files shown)


src/PAModel/CanvasDocument.cs
```diff
@@ -351,17 +351,23 @@ internal void ApplyAfterMsAppLoadTransforms(ErrorContainer errors)
 
             // Shard templates, parse for default values
             var templateDefaults = new Dictionary<string, ControlTemplate>();
-            foreach (var template in _templates.UsedTemplates)
```
NavneetThekkumpat, line 354: You could make this simpler if nullable by using the question mark to check for null
 `foreach (var template in _templates?.UsedTemplates)`

NavneetThekkumpat: Same suggestion in almost all checks below

tomjschneider:
Do we have tests for any of these scenarios? Are these hypothetical or things we can reproduce?

petrochuk:
Old PR with conflicts

PR #466, opened 2023-02-27
Handling PCF template data separately for each PCF control instance

NavneetThekkumpat:
## Problem
- Different control instances could have different template data even if they appear to follow same template.
- Like version, DynamicControlDefinitionJSON etc could be different.
``` 
Eg: Control Instance 1 -> pcftemplate1 -> DynamicControlDefinitionJson1
      Control Instance 2 -> pcftemplate1 -> DynamicControlDefinitionJson2
```
- This causes roundtrip errors like below
image

## Solution
- Make use of entropy data and create a new dictionary to store/retrieve PCF template data
- Store PCFtemplate separately while unpacking >>> ` IRStateHelpers.SplitIRAndState()`
- Retrieve the stored PCFtemplate data while packing >>> `IRStateHelpers.CombineIRAndState()`
- This will avoid such roundtrip errors

## Validation
- UnitTesting

### DIFF
--- a/src/PAModel/IR/IRStateHelpers.cs
+++ b/src/PAModel/IR/IRStateHelpers.cs
@@ -10,6 +10,7 @@ using System.Linq;
 using Microsoft.AppMagic.Authoring.Persistence;
 using System;
 using Microsoft.PowerPlatform.Formulas.Tools.Schemas;
+using static Microsoft.PowerPlatform.Formulas.Tools.ControlInfoJson;
 
 namespace Microsoft.PowerPlatform.Formulas.Tools
 {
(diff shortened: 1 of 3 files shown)


src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -196,6 +197,15 @@ private static void SplitIRAndState(ControlInfoJson.Item control, string topPare
                 entropy.OverridablePropertiesEntry.Add(control.Name, OverridablePropVal);
             }
 
+            // Storing pcf controls template data in entropy
+            // Since this could be different for different controls, even if that appear to follow the same template
+            // Eg:Control Instance 1 -> pcftemplate1 -> DynamicControlDefinitionJson1
+            // Control Instance 2 -> pcftemplate1 -> DynamicControlDefinitionJson2
+            if (IsPCFControl(control.Template))
+            {
+                entropy.PCFTemplateEntry.Add(control.Name, control.Template);
```
tomjschneider, line 206: Just to double check - is this template being stored anywhere else today? Will it be stored in both this `PCFTemplateEntry` collection and another?

tomjschneider (approved):

src/PAModel/IR/IRStateHelpers.cs
NavneetThekkumpat: That is a good point. Let me try removing from the other collection (templatestore) and check if no sideeffects.

tomjschneider: I think you still need template store for components and controls though, right?

tomjschneider (approved):

PR #473, opened 2023-05-15
Handling host control instances with template hostype/hostservice values

NavneetThekkumpat:
## Problem
Host control template hostype/hostservice values are newly introduced and not supported by PALT. We need to store these values unchanged while unpacking/packing.

Host control is a singleton control added to every single msapp. But integration controls for apps that exists in embedded scenarios like sharepointintegration/modeldrivenformintegration are also recognized as **host controls.**

1. To repro multiple host control instances with different **HostType**, try creating an integrated SharePoint powerapps(like the one attached to this PR code changes).

1. One example of host control instances with **HostService** values would be a canvas app from modeldrivenformintegration  https://learn.microsoft.com/en-us/power-apps/maker/model-driven-apps/embedded-canvas-app-properties-actions 

## Solution

Template values could be different for each host control instances.
Considering that, we need to store each of these template values separately in templatestore, rather than once for hostcontrol. 
This enables Storing Template HostType and HostService details for each host control instances.
 Eg:
 ``` 
Host Control Instance 1 -> template1 -> HostType1
Host Control Instance 2 -> template1 -> HostType2
```
OR
```
   Host Control Instance 1 -> template1 -> HostService1
   Host Control Instance 2 -> template1 -> HostService2
```

## Validation
- .msapp roundtrips well
- Unittesting

### CODE REVIEW

**Review by landonarogers — APPROVED** [review]

**Comment by tomjschneider** [comment]
What happens if entropy is cleared out like we saw with the PCF templates?

**Comment by NavneetThekkumpat** [comment]
> What happens if entropy is cleared out like we saw with the PCF templates?

I shall test this scenario. Working on unittesting

File: src/PAModelTests/EntropyTests.cs [hunk 1]
```diff
@@ -154,5 +155,101 @@ public void TestPCFControlWillFallBackToControlTemplate(string appName)
...
+            using (var tempFile = new TempFile())
+            {
+                MsAppSerializer.SaveAsMsApp(msapp, tempFile.FullPath, new ErrorContainer());
+                Assert.IsTrue(MsAppTest.Compare(root, tempFile.FullPath, Console.Out));
+            }
+        }
+
+        [DataTestMethod]
+        [DataRow("HostControlTestWithHostTypeAndHostService.msapp")]
+        public void TestHostControlWithHostTypeAndServiceFallBackToTemplate(string appName)
```
tomjschneider, line 229: nit: Is this test name (and the one above it) accurate? It doesn't fall back to the template if it is missing from Entropy, does it?

src/PAModelTests/EntropyTests.cs
```diff
@@ -154,5 +155,101 @@ public void TestPCFControlWillFallBackToControlTemplate(string appName)
...
+
+            Assert.AreEqual("Host", sourceFile.Value.TopParent.Children.Last().Name);
+            // Checking if the HostType and HostService Entries were added
+            Assert.IsTrue(sourceFile.Value.TopParent.Children.Last().Template.ExtensionData.ContainsKey(IRStateHelpers.ControlTemplateHostTypePropertyName));
+            Assert.IsTrue(sourceFile.Value.TopParent.Children.Last().Template.ExtensionData.ContainsKey(IRStateHelpers.ControlTemplateHostServicePropertyName));
+        }
+
+        [DataTestMethod]
+        [DataRow("HostControlTestWithHostType.msapp")]
+        public void TestHostControlWithHostTypeFallBackToTemplate(string appName)
```
tomjschneider, line 205: Do these tests fail before your fix? Just looking at this, I'm not sure I understand how these tests would be passing since the entropy data was the only place this info was stored, and it has been cleared out.

NavneetThekkumpat: Good catch! Okay, so looks like this app needs some work. Let me look into it.

src/PAModelTests/EntropyTests.cs
NavneetThekkumpat: You are right it does not fallback these values needs to be stored especially when different control instances are having the same host control template. 
Changed the unit tests to match this.

src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -11,13 +11,17 @@
 using System;
 using Microsoft.PowerPlatform.Formulas.Tools.Schemas;
 using static Microsoft.PowerPlatform.Formulas.Tools.ControlInfoJson;
+using System.Drawing;
```
tomjschneider, line 14: I don't think you meant to have these two new `using` statements?

src/PAModelTests/EntropyTests.cs
```diff
@@ -154,5 +155,105 @@ public void TestPCFControlWillFallBackToControlTemplate(string appName)
...
+            Assert.IsTrue(msapp._entropy.HostTypeEntry.Count == 0);
+
+            msapp._entropy.HostServiceEntry.Clear();
+            Assert.IsTrue(msapp._entropy.HostServiceEntry.Count == 0);
+
+            // Repack the app and validate it does not match the initial msapp
+            using (var tempFile = new TempFile())
+            {
+                MsAppSerializer.SaveAsMsApp(msapp, tempFile.FullPath, new ErrorContainer());
+                Assert.IsFalse(MsAppTest.Compare(root, tempFile.FullPath, Console.Out));
```
tomjschneider, line 255: I'm a bit confused here. Why would we want a unit test for an app not successfully unpacking?

tomjschneider (approved):
Discussed with Navaneeth - we believe that storing this in Entropy is fine because this data being cleared out would not impact the actual loading of an `.msapp`. Navaneeth is going to confirm with the team.

src/PAModel/IR/IRStateHelpers.cs
```diff
@@ -18,6 +18,8 @@ internal static class IRStateHelpers
     {
         public const string ControlTemplateOverridableProperties = "OverridableProperties";
         public const string ControlTemplatePCFDynamicSchemaForIRRetrieval = "PCFDynamicSchemaForIRRetrieval";
+        public const string ControlTemplateHostTypePropertyName = "HostType";
```
tomjschneider, line 21: Is this still used anywhere? I only see it used in the unit test.

tomjschneider (approved):

src/PAModel/IR/IRStateHelpers.cs
NavneetThekkumpat: You are right, used only in unit test.

Issue #479, opened 2023-06-15
'pac solution clone' failing with 'PA3015: Property Removed' error 


RajeevPentyala:
- [ ] Please attach a sample .msapp to this Issue which demonstrates the issue you are experiencing. *(Please do not attach any file which may contain sensitive personal or company data.)*

**Describe the bug**

I've a 'Dataverse' solution with a 'Page' component. While cloning the solution using 'pac solution clone' command, getting following error.

_Error: Error   PA3015: Property Removed: LocalDatabaseReferences.default.cds.dataSources.environment_39a902ba.entitySetName
Error   PA3015: Property Removed: LocalDatabaseReferences.default.cds.dataSources.environment_39a902ba.logicalName
Error   PA3011: Roundtrip validation on unpack failed._

**To Reproduce**

Clone the solution using 'pac solution clone' command.

**Screenshots**

![image](https://github.com/microsoft/PowerApps-Language-Tooling/assets/104883923/2ab02227-c529-4db9-b16e-f2b12b79930b)

**Additional context**

Solution has a 'page' component.

akshar2401:
Pr linked with this issue got merged recently and would ultimately fix this issue. The fix should be shipped in the next release of PAC CLI

Issue #475, opened 2023-05-22
PA3015: Property Removed, Edited app, dataSources.environment


landonarogers:
Here to also report the "PA3015:Property Removed" errors.  We have one specific project which fails with those errors.  Uncertain what the exact cause was on our side, but what @PatrickMichaelKriegler reported would make sense given the context.  I unfortunately do not have permission to share our msapp files. 

2023-05-09T15:35:57.6725991Z   source format version: 0.24
2023-05-09T15:36:04.6877128Z Warning PA2001: Checksum mismatch. Checksum indicates that sources have been edited since they were unpacked. If this was intentional, ignore this warning.
2023-05-09T15:36:04.7533088Z Microsoft PowerPlatform CLI
2023-05-09T15:36:04.7535526Z Version: 1.22.3+g8af83fd
2023-05-09T15:36:04.7536067Z 
2023-05-09T15:36:04.7542375Z Error: Error   PA3015: Property Removed: LocalDatabaseReferences.default.cds.dataSources.environment_39a902ba.entitySetName
2023-05-09T15:36:04.7542727Z Error   PA3015: Property Removed: LocalDatabaseReferences.default.cds.dataSources.environment_39a902ba.logicalName
2023-05-09T15:36:04.7543073Z Error   PA3011: Roundtrip validation on unpack failed. 
2023-05-09T15:36:04.7543566Z You have found a bug; this is not a specific bug, rather an indicator that some bug has been encountered.
2023-05-09T15:36:04.7544168Z Please open an issue and log the entirety of this error log at https://github.com/microsoft/PowerApps-Language-Tooling
2023-05-09T15:36:04.7544359Z 
2023-05-09T15:36:04.7544460Z 
2023-05-09T15:36:04.7544709Z     Error   PA3015: Property Removed: LocalDatabaseReferences.default.cds.dataSources.environment_39a902ba.entitySetName
2023-05-09T15:36:04.7545049Z Error   PA3015: Property Removed: LocalDatabaseReferences.default.cds.dataSources.environment_39a902ba.logicalName
2023-05-09T15:36:04.7545333Z Error   PA3011: Roundtrip validation on unpack failed. 
2023-05-09T15:36:04.7545609Z You have found a bug; this is not a specific bug, rather an indicator that some bug has been encountered.
2023-05-09T15:36:04.7546021Z Please open an issue and log the entirety of this error log at https://github.com/microsoft/PowerApps-Language-Tooling

[...]

PR #484, opened 2023-07-13
Preventing null ref and null hash mismatch and property removed with LocalDatabaseReferenceJson.dataSources

akshar2401:
https://github.com/microsoft/PowerApps-Language-Tooling/issues/479
https://github.com/microsoft/PowerApps-Language-Tooling/issues/475

## Problem

This pr attempts to fix the "Property Removed" errors and dataSources being null exception.. Additionally, this pr fixes some of the null ref issues i found while testing. After fixing null ref for dataSources, saw a new hash mismatch error caused by dataSources being null in original msapp but being written as {} in an msapp generated after repacking the sources created from the original msapp.
image

## Solution

 It does so by tracking unused data sources in all data source definitions files (in pkgs/TableDefinitions). that are generated for each data source for a given local database reference. When reading back the sources, it picks up the unused data sources from the first data source definition file for that given local database refereence. The solution to track unused datasources also fixes the issue which caused dataSources to be written as {} in msapp generated after repacking the sources created from the original msapp which had a null value for dataSources

## Validation

-Testing


src/PAModelTests/DataSourceTests.cs
```diff
@@ -117,6 +120,142 @@ public void TestConnectionInstanceIDHandling(string appName)
...
+            var sourcesTempDirPath = sourcesTempDir.Dir;
+            errors = msApp.SaveToSources(sourcesTempDirPath, pathToMsApp);
+            (msApp, errors) = CanvasDocument.LoadFromSources(sourcesTempDirPath);
+            using var msAppTemp = new TempFile();
+            using var sources2 = new TempDir();
+            var errors2 = new ErrorContainer();
+            MsAppSerializer.SaveAsMsApp(msApp, msAppTemp.FullPath, errors2);
+            errors = msApp.SaveToSources(sources2.Dir, msAppTemp.FullPath);
+            errors.ThrowOnErrors();
+            var entropy = ReadEntropy(sources2.Dir);
```
NavneetThekkumpat, line 237: Need to fix the file not found error here
image

src/PAModelTests/DataSourceTests.cs
```diff
@@ -117,6 +120,142 @@ public void TestConnectionInstanceIDHandling(string appName)
...
+            var (msApp, errors) = CanvasDocument.LoadFromMsapp(pathToMsApp);
+            msApp._dataSourceReferences["default.cds"].dataSources = null;
+            errors.ThrowOnErrors();
+
+            using var sourcesTempDir = new TempDir();
+            var sourcesTempDirPath = sourcesTempDir.Dir;
+            errors = msApp.SaveToSources(sourcesTempDirPath, pathToMsApp);
+            (msApp, errors) = CanvasDocument.LoadFromSources(sourcesTempDirPath);
+            using var msAppTemp = new TempFile();
+            using var sources2 = new TempDir();
```
NavneetThekkumpat, line 232: Maybe make use of
 ``` 
 using (var tempDir = new TempDir())
 
Example usage PowerApps-Language-Tooling\src\PAModelTests\AppTestsTest.cs

NavneetThekkumpat (approved):

PR #486, opened 2023-07-19
Fixed multiline msapp pack errors

petrochuk:

## Problem

```
Error: Error   PA3013: Property Value Changed: TopParent.Children[0].Children[1].Rules[2].InvariantScript
Error   PA3011: Roundtrip validation on unpack failed.
You have found a bug; this is not a specific bug, rather an indicator that some bug has been encountered.
Please open an issue and log the entirety of this error log at https://github.com/microsoft/PowerApps-Language-Tooling
```

## Solution

While validating we use normalized line endings but YamlLexer has Windows style endings hardcoded

## Validation

Updated unit tests to validate

### DIFF
--- a/src/PAModel/PAConvert/Yaml/YamlLexer.cs
+++ b/src/PAModel/PAConvert/Yaml/YamlLexer.cs
@@ -119,7 +119,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools.Yaml
     /// </summary>
     internal class YamlLexer
     {
-        private static string NewLine = "\r\n";
+        private static string NewLine = "\n";
 
         // The actual contents to read. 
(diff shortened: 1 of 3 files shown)


swapnilbhalgat (approved):

tomjschneider (approved):

tomjschneider:
It looks like the message we create in `src\PAModel\PAConvert\ErrorCode.cs` for an `InternalError` also contains Windows style endings, if you want to fix that while you're in here as well. Not required.

tomjschneider (commented):
It looks like the message we create in src\PAModel\PAConvert\ErrorCode.cs for an InternalError also contains Windows style endings, if you want to fix that while you're in here as well. Not required.

PR #488, opened 2023-07-27
Fix Missing Logo File Throwing an Error

landonarogers:
## Problem

We don't want a missing logo file to throw an error, we'd like it to roundtrip so as not to block the user from packing the app.

## Solution

Remove the error that throws, and make sure the FileName is not null.

## Validation

- Self validation

### DIFF
--- a/src/PAModel/Serializers/SourceSerializer.cs
+++ b/src/PAModel/Serializers/SourceSerializer.cs
@@ -368,10 +368,6 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
                     app._unknownFiles.Remove(key);
                     app._logoFile = logoFile;
                 }
-                else
-                {
-                    throw new InvalidOperationException($"Missing logo file {key}");
-                }
             }
         }
 
--- a/src/PAModel/Serializers/TransformLogo.cs
+++ b/src/PAModel/Serializers/TransformLogo.cs
@@ -48,7 +48,7 @@ namespace Microsoft.PowerPlatform.Formulas.Tools
             FileEntry logoFile = null;
             var publishInfo = app._publishInfo.JsonClone();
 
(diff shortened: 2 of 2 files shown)


src/PAModel/Serializers/SourceSerializer.cs
```diff
@@ -368,10 +368,6 @@ private static void GetLogoFile(this CanvasDocument app)
                     app._unknownFiles.Remove(key);
                     app._logoFile = logoFile;
                 }
-                else
-                {
-                    throw new InvalidOperationException($"Missing logo file {key}");
```
jt000, line 373: You may want to consider logging something to our debug logs in this scenario to make sure we're aware of the odd scenario if we get a CRI in the future...

landonarogers: No internal debug logging for this repo

src/PAModel/Serializers/TransformLogo.cs
```diff
@@ -48,7 +48,7 @@ public static (PublishInfoJson, FileEntry) TransformLogoOnSave(this CanvasDocume
             FileEntry logoFile = null;
             var publishInfo = app._publishInfo.JsonClone();
 
-            if (!string.IsNullOrEmpty(publishInfo?.LogoFileName))
+            if (!string.IsNullOrEmpty(publishInfo?.LogoFileName) && app._logoFile?.Name != null)
```
tomjschneider, line 51: Should `app._logoFile?.Name != null` is a `string.IsNullOrEmpty` check?

tomjschneider, line 51: One potential gap I see in this new logic: what if `app._entropy.OldLogoFileName` has a value? Previously that would be used and preferred over the `app._logoFile.Name`, now it will be skipped. Is that desired?

tomjschneider (approved):

PR #495, opened 2023-09-14
Add analysis info to header for yaml, fix missing template information in packed .msapp

jack-work:
Fixes this bug present due to newly introduced .msapp properties.
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -357,3 +357,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# Ignore .idea IDE files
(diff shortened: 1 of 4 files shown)


akshar2401 (approved):

petrochuk (approved):

jt000 (changes requested):
Please abandon, this issue was previously fixed in https://github.com/microsoft/PowerApps-Language-Tooling/commit/941f06c8f5cf7a493289e9514c8dec5c50b61d34 in a similar way that was mentioned in your comments. Please let me know if you're still able to repro after pulling this commit.

PR #516, opened 2024-01-24
updated project signing config

mizrael:
updated POCOs project signing config


src/Microsoft.PowerPlatform.PowerApps.Persistence.Tests/Microsoft.PowerPlatform.PowerApps.Persistence.Tests.csproj
```diff
@@ -7,6 +7,9 @@
 
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
+    <SignAssembly>true</SignAssembly>
```
petrochuk, line 10: Signing is slow. let's skip it for tests

petrochuk (approved):

tehcrashxor:
We'll also need `azure-pipelines.yml`'s signing paths updated at some point

mizrael:
> We'll also need `azure-pipelines.yml`'s signing paths updated at some point

was going to do this right after, but that will need some trial and error

PR #517, opened 2024-01-25
Using same YamlDotNet version

petrochuk:


src/Microsoft.PowerPlatform.PowerApps.Persistence/Microsoft.PowerPlatform.PowerApps.Persistence.csproj
```diff
@@ -46,7 +46,7 @@
     </PackageReference>
   </ItemGroup>
 
-  <Target Name="PALT" AfterTargets="MinVer">
+  <Target Name="SetAssemblyVersion" AfterTargets="MinVer">
```
tehcrashxor, line 49: Should probably update the name for this task in the other csproj as well

tehcrashxor (approved):

PR #521, opened 2024-01-25
updated az pipeline

mizrael:
updated the AZ pipeline to include the new Persistence project.


src/PASopa_pack.sln
```diff
@@ -0,0 +1,36 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
petrochuk, line 2: Why do we need separate solution file?

azure-pipelines.yml
```diff
@@ -113,6 +113,57 @@ jobs:
         ]
       ServiceEndpointUrl: 'https://api.esrp.microsoft.com/api/v1'
 
+  - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@3
```
petrochuk, line 116: Can we use just one ESRP task?

src/PASopa_pack.sln
mizrael: this includes only the projects that need to be published as nuget packages, avoids running the pack job for each one

azure-pipelines.yml
mizrael: haven't found a decent way to do that (yet)

mizrael: yes, just updated the code

tehcrashxor (approved):

src/PASopa_pack.sln
petrochuk: let's see if we can avoid extra solution

mizrael: added an array parameter to the `pack` step to specify the projects to include

PR #525, opened 2024-01-25
serializing control properties in alphabetical order

mizrael:
serializing control properties in alphabetical order

## Validation

Unit tests

### DIFF
--- a/src/Microsoft.PowerPlatform.PowerApps.Persistence/Models/Text.cs
+++ b/src/Microsoft.PowerPlatform.PowerApps.Persistence/Models/Text.cs
@@ -5,11 +5,11 @@ using Microsoft.PowerPlatform.PowerApps.Persistence.Attributes;
 
 namespace Microsoft.PowerPlatform.PowerApps.Persistence.Models;
 
-[FirstClass(nameof(Label))]
-public record Label : Control
+[FirstClass(nameof(Text))]
+public record Text : Control
 {
-    public Label()
+    public Text()
     {
-        ControlUri = BuiltInTemplatesUris.Label;
+        ControlUri = BuiltInTemplatesUris.Text;
     }
 }
--- a/src/Microsoft.PowerPlatform.PowerApps.Persistence/Yaml/ControlPropertiesCollectionConverter.cs
+++ b/src/Microsoft.PowerPlatform.PowerApps.Persistence/Yaml/ControlPropertiesCollectionConverter.cs
@@ -41,7 +41,7 @@ public class ControlPropertiesCollectionConverter : IYamlTypeConverter
 
         emitter.Emit(new MappingStart());
 
-        var sortedKeys = collection.Keys.OrderBy(x => x);
+        var sortedKeys = collection.Keys.OrderBy(x => x, StringComparer.Ordinal);
 
         foreach (var key in sortedKeys)
         {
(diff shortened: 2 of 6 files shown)


tehcrashxor (approved):

src/Microsoft.PowerPlatform.PowerApps.Persistence/Models/Text.cs
```diff
@@ -5,11 +5,11 @@
 
 namespace Microsoft.PowerPlatform.PowerApps.Persistence.Models;
 
-[FirstClass(nameof(Label))]
-public record Label : Control
```
petrochuk, line 9: Why did you change from "label" to "text"? Just trying to follow the logic

petrochuk (commented):
Maybe later but can we also add deserialization tests for actual yaml files?

src/Microsoft.PowerPlatform.PowerApps.Persistence/Models/Text.cs
mizrael: renamed in favor of the actual control we have on the portal

src/Microsoft.PowerPlatform.PowerApps.Persistence/Yaml/ControlPropertiesCollectionConverter.cs
```diff
@@ -41,12 +41,15 @@ public void WriteYaml(IEmitter emitter, object? value, Type type)
 
         emitter.Emit(new MappingStart());
 
-        foreach (var kv in collection)
+        var sortedKeys = collection.Keys.OrderBy(x => x);
```
petrochuk, line 44: Should we specify culture for sorting?

src/Microsoft.PowerPlatform.PowerApps.Persistence/Yaml/SerializerBuilderExtensions.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.PowerPlatform.PowerApps.Persistence.Yaml;
 
 internal static class SerializerBuilderExtensions
 {
+    private const string ControlPropertyName = "Control";
```
petrochuk, line 13: Probably builder extension is not the best place for this constant

mizrael: agreed. But we're not referencing it anywhere else, and we use "Control" only when serializing into YAML for now. I'd move it outside when we need to

src/Microsoft.PowerPlatform.PowerApps.Persistence/Yaml/ControlPropertiesCollectionConverter.cs
mizrael: good catch. `StringComparer.CurrentCultureIgnoreCase` ?

tehcrashxor: Yeah, probably an Ordinal comparison.
I'm not sure what DocumentServer is doing for culture settings, but there could very well be a per-thread culture in response to requests.

petrochuk (approved):

PR #531, opened 2024-01-29
refactoring

mizrael:
Removed YAML references from the `Control` class to make it persistence-agnostic.
General refactoring and code cleanup.


src/Persistence/Yaml/ControlTypeDiscriminator.cs
```diff
@@ -16,11 +16,11 @@ public bool TryDiscriminate(IParser buffer, out Type? suggestedType)
     {
         suggestedType = null;
 
-        if (!buffer.TryFindMappingEntry(s => BuiltInTemplates.ShortNameToType.ContainsKey(s.Value) || s.Value == Control.YamlName, out var scalar, out var value))
+        if (!buffer.TryFindMappingEntry(s => BuiltInTemplates.ShortNameToType.ContainsKey(s.Value) || s.Value == nameof(Control), out var scalar, out var value))
             return false;
 
         // Control is abstract, so we need to return a concrete type.
-        if (scalar!.Value == Control.YamlName)
```
petrochuk, line 23: Class name and string used in Yaml can be different

src/Persistence/Utils/FileUtils.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.Utils;
+
+public static class FileUtils
```
petrochuk, line 6: Please avoid Utils pattern

src/Persistence/MsappArchive.cs
```diff
@@ -317,4 +286,15 @@ public void Dispose()
     }
 
     #endregion
+
+    public static class Directories
+    {
+        public const string SrcDirectory = "Src";
```
petrochuk, line 292: If class name is Directories. There is no need to add "Directory" to each field

src/Persistence/Yaml/SerializerBuilderExtensions.cs
```diff
@@ -32,4 +33,21 @@ public static DeserializerBuilder WithFirstClassModels(this DeserializerBuilder
             })
             .WithTypeConverter(new ControlPropertiesCollectionConverter());
     }
+
+    private static TBuilder AddAttributeOverrides<TBuilder>(this TBuilder builder)
+       where TBuilder : BuilderSkeleton<TBuilder>
+    {
+        var types = typeof(Control).Assembly.DefinedTypes;
+        foreach (var type in types)
```
petrochuk, line 41: Why do we still need this?

src/Persistence/Yaml/YamlUtils.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.Yaml;
+
+public static class YamlUtils
```
petrochuk, line 6: Please avoid utils pattern

petrochuk (changes requested):

src/Persistence/MsappArchive.cs
mizrael: right, I was going to do that too!

src/Persistence/Yaml/SerializerBuilderExtensions.cs
mizrael: this helps keeping the `Control` class persistence-agnostic, so we don't pollute it with attributes specific to the file format (eg. YAML)

src/Persistence/Yaml/YamlUtils.cs
mizrael: I removed the Screens and the EditorState deserialization for now. Let's discuss during standup.

src/Persistence.Tests/MsApp/MsappArchiveTests.cs
```diff
@@ -64,28 +65,11 @@ public void AddEntryTests(string[] entries)
...
         }
 
         // Get the required entry should throw if it doesn't exist
         var action = () => msappArchive.GetRequiredEntry("not-exist");
         action.Invoking(a => a()).Should().Throw<FileNotFoundException>();
     }
-
-    [TestMethod]
-    [DataRow(@"_TestData/AppsWithYaml/HelloWorld.msapp", 14, 2, "HelloScreen", "screen", 8)]
-    public void Msapp_ShouldHave_Screens(string testFile, int allEntriesCount, int controlsCount,
```
petrochuk, line 77: Please do not delete unit tests

src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
```diff
@@ -10,12 +9,12 @@ namespace Persistence.Tests.Yaml;
...
 {
     [TestMethod]
-    public void Deserialize_ShouldFail()
+    public void Deserialize_ShouldFailWhenYamlIsInvalid()
     {
         // Arrange
         var deserializer = YamlSerializationFactory.CreateDeserializer();
 
-        var files = Directory.GetFiles(@"_TestData/InvalidYaml", $"*{MsappArchive.YamlFxFileExtension}", SearchOption.AllDirectories);
+        var files = Directory.GetFiles(@"_TestData/InvalidYaml", $"*.fx.yaml", SearchOption.AllDirectories);
```
petrochuk, line 17: Why do we need to hard code extension here?

src/Persistence/Extensions/IMsappArchiveExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Returns the entry in the archive with the given name or throws a <see cref="FileNotFoundException"/> if it does not exist.
+    /// </summary>
+    /// <param name="archive">the <see cref="IMsappArchive"/> instance.</param>
+    /// <param name="entryName">the name of the entry to fetch.</param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException"></exception>
+    /// <exception cref="FileNotFoundException"></exception>
+    public static ZipArchiveEntry GetRequiredEntry(this IMsappArchive archive, string entryName)
```
petrochuk, line 18: This can be part of MsappArchive class

src/Persistence/Models/Control.cs
```diff
@@ -50,15 +45,10 @@ public Control(string name)
...
     public ControlPropertiesCollection Properties { get; init; } = new();
 
     /// <summary>
     /// list of child controls nested under this control.
-    /// </summary>
-    [YamlMember(Order = 3)]
-    public IList<Control> Controls { get; init; } = new List<Control>();
-
-    [YamlIgnore]
-    public ControlEditorState? EditorState { get; set; }
```
petrochuk, line 63: It is marked as YamlIngore. why do you need to delete it?

src/Persistence/Models/ControlEditorState.cs
petrochuk, line 1: Why deleting this file?

src/Persistence/Utils/FileUtils.cs
```diff
@@ -0,0 +1,17 @@
...
+internal static class FileUtils
+{
+    /// <summary>
+    /// Converts backslashes to forward slashes, removes trailing slashes, and converts to lowercase.
+    /// </summary>
+    /// <param name="path"></param>
+    /// <returns></returns>
+    public static string NormalizePath(string path)
+    {
+        return path.Trim().Replace('\\', '/').Trim('/').ToLowerInvariant();
```
petrochuk, line 15: This normalize pattern is specific to Zip entries

src/Persistence/Yaml/SerializerBuilderExtensions.cs
petrochuk: I don't think adding attributes is polluting. I am thinking injecting attributes at runtime is very unexpected behavior

petrochuk (changes requested):

src/Persistence.Tests/MsApp/MsappArchiveTests.cs
mizrael: I dropped the code that loads Screens from msapp, this testt is not necessary anymore. We have the same thing already in Docserver. Will eventually migrate that over here in time

src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
mizrael: I removed the `YamlUtils` class along with the `Screen` loading code from msapp.

src/Persistence/Models/Control.cs
mizrael: we're not loading Editor state here for the moment

src/Persistence/Models/ControlEditorState.cs
mizrael: we discussed during standup, we're not loading and merging editor state in this project at the moment, this will be done in a future PR once the schema and POCOs are stable

src/Persistence/Models/Control.cs
```diff
@@ -50,15 +45,10 @@ public Control(string name)
...
+    /// </summary>    
+    public Control[] Controls { get; init; } = Array.Empty<Control>();
```
tehcrashxor, line 53: Perhaps out of scope of this PR:

Apps, Screens, and Canvas Components have child objects (Canvas Components and Screens both contain controls, Apps contain Screens and Canvas Components), but to my knowledge no other controls can have child controls.  Perhaps this property should be moved either directly to those types, or to somewhere else in the inheritance hierarchy.

src/Persistence/Yaml/SerializerBuilderExtensions.cs
mizrael: it's more declarative. EntityFramework does something similar when building the config for the entities definition. 
The real problem that we want to solve is avoiding having configuration/attributes specific to a file format (YAML in our case) embedded into our POCO classes. It would lead to the same problem we have right now on DocServer with JSON

src/Persistence/Utils/FileUtils.cs
mizrael: updated the code, managed to remove the reference to this method from the tests project

src/Persistence/Models/Control.cs
mizrael: this is a good point. We're basically saying that only these three types of objects can have a tree-like structure (`Apps`, `Screens` and `Components` ). Everything else is flat.
Let's discuss with the architects first, I want to make sure we're covering all the scenarios. My point is that removing the ability to create a tree might backfire in the future and I want to avoid that.

src/Persistence/Yaml/SerializerBuilderExtensions.cs
petrochuk: Control class is designed for persistence. It is only intended to hold data during read/write operations.

mizrael: that's correct, but we should avoid having details specific to the file format in these classes

petrochuk (approved):

PR #532, opened 2024-01-30
Added reading of top level screens by MsAppArchive

petrochuk:


mizrael (changes requested):
can we hold on on this? It might conflict with the changes I will be adding on Docserver. Can we have a meeting first and go over the design?

petrochuk:
> can we hold on on this? It might conflict with the changes I will be adding on Docserver. Can we have a meeting first and go over the design?

we can but this is adding new reading functionality. Let's chat offline

mizrael (changes requested):
see my comments

PR #534, opened 2024-01-30
skipping unmatched properties during deserialization

mizrael:


src/Persistence/Yaml/SerializerBuilderExtensions.cs
```diff
@@ -24,6 +24,7 @@ public static SerializerBuilder WithFirstClassModels(this SerializerBuilder buil
     public static DeserializerBuilder WithFirstClassModels(this DeserializerBuilder builder)
     {
         return builder
+            .IgnoreUnmatchedProperties()
```
petrochuk, line 27: I can approve for now but we need to discuss this

petrochuk (approved):

src/Persistence/Yaml/SerializerBuilderExtensions.cs
petrochuk: the issue here is if customer makes mistake in Yaml we will quietly ignore it instead of reporting it as an error

PR #542, opened 2024-02-05
Added optionall controls parameter

petrochuk:


src/Persistence/Templates/ControlFactory.cs
```diff
@@ -15,14 +15,22 @@ public ControlFactory(IControlTemplateStore controlTemplateStore)
         _controlTemplateStore = controlTemplateStore ?? throw new ArgumentNullException(nameof(controlTemplateStore));
     }
 
-    public Control Create(string name, string template, ControlPropertiesCollection? properties = null)
+    public Control Create(string name, string template, ControlPropertiesCollection? properties = null, Control[]? controls = null)
```
mizrael, line 18: nit: instead an array, it would be better to use `IEnumerable<>`. It's good practice to be more relaxed on input types and more specific on return types.

petrochuk: https://github.com/microsoft/PowerApps-Tooling/pull/543

PR #543, opened 2024-02-05
Converted from array to IEnumerable

petrochuk:


src/Persistence/Templates/ControlFactory.cs
```diff
@@ -15,21 +15,21 @@ public ControlFactory(IControlTemplateStore controlTemplateStore)
...
-                Controls = controls ?? Array.Empty<Control>()
+                Controls = (Control[])(controls ?? Array.Empty<Control>())
             };
         }
 
         return new CustomControl(name, new ControlTemplate(template))
         {
             Properties = properties ?? new(),
-            Controls = controls ?? Array.Empty<Control>()
+            Controls = (Control[])(controls ?? Array.Empty<Control>())
```
mizrael, line 32: I suspect that cast might fail if `controls` is something different than array. This should be safer:
`(controls ?? Array.Empty<Control>()).ToArray()`

mizrael (changes requested):

src/Persistence/Templates/ControlFactory.cs
petrochuk: your suggestion will create a new copy of array

mizrael: if `controls` is a List or any other collection type, the explicit cast to `Control[]` will throw. If we want to reduce allocations then we should check if it's an array before casting. Something like this:
```
Control[] tmpArray = null;
 if (controls is Control[] arr)
     tmpArray = arr;
 else
     tmpArray = (controls ?? Array.Empty<Control>()).ToArray();
 ```

mizrael (changes requested):

PR #553, opened 2024-02-08
Added Update archive methods

petrochuk:


src/Persistence/MsApp/IMsappArchiveFactory.cs
```diff
@@ -19,12 +19,22 @@ public interface IMsappArchiveFactory
...
-    /// Opens existing msapp archive.
+    /// Opens existing msapp archive for read.
     /// </summary>
     IMsappArchive Open(string path);
 
     /// <summary>
-    /// Opens existing msapp archive.
+    /// Opens existing msapp archive for read.
     /// </summary>
     IMsappArchive Open(Stream stream, bool leaveOpen = false);
```
mizrael, line 29: we might start considering having 2 separate interfaces, one for reading and another one for writing. The `MsappArchive` implementation can very well implement both (and we keep it internal), but from the factory instead we return either an `IMappArchiveReader` or a writer. This would help keeping the interfaces small and avoid cases where a client opens an archive for reading and then tries to write too

petrochuk: Good idea for future improvements. At this time if client opens for reading and tries to write there will be an exception coming from ZipArchive

mizrael (approved):

PR #569, opened 2024-03-01
micro optimization

mizrael:
image

with .NET 8 we could gain a little bit more by passing params by ref

### DIFF
--- a/src/Persistence.Tests/Extensions/StringExtensionsTest.cs
+++ b/src/Persistence.Tests/Extensions/StringExtensionsTest.cs
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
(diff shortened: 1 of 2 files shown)


petrochuk (approved):

src/Persistence.Tests/Extensions/StringExtensionsTest.cs
```diff
@@ -0,0 +1,44 @@
...
+#pragma warning disable CS8604
+
+[TestClass]
+public class StringExtensionsTest
+{
+    [TestMethod]
+    public void FirstCharToUpper_NullInput_ReturnsNull()
+    {
+        string input = null;
+        Assert.IsNull(input.FirstCharToUpper());
```
joem-msft, line 18: this should throw.
Extension methods should be validating that the `this` parameter is not null before using it.
Calling code should use `input?.FirstCharToUpper()` when input could be null.

PR #573, opened 2024-03-05
Add support for msapp generation commandline arguments

cwduncan:
## Problem

The initial implementation of Test.AppWriter can only generate a static example app

## Solution

Adding more in depth commandline argument support for MSApp generation in Test.AppWriter

## Changes

- Added support for System.Commandline to support more robust commandline arguments
- Implemented initial arguments to support the filepath, number of screens, and a list of controls

## Validation

### DIFF
--- a/samples/Test.AppWriter/ExampleAppGenerator.cs
+++ b/samples/Test.AppWriter/ExampleAppGenerator.cs
@@ -1,11 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-//using System;
-//using System.Collections.Generic;
-//using System.Linq;
-//using System.Text;
-//using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.PowerPlatform.PowerApps.Persistence.Templates;
 using Microsoft.PowerPlatform.PowerApps.Persistence.Models;
--- a/samples/Test.AppWriter/ExampleAppGenerator.cs
+++ b/samples/Test.AppWriter/ExampleAppGenerator.cs
@@ -15,13 +15,17 @@ internal class ExampleAppGenerator
 {
     // This produces a simple example app including the specified number of screens and a few generic controls
     // This is intended for testing purposes only
-    public static App GetExampleApp(IServiceProvider provider, string appName)
+    public static App GetExampleApp(IServiceProvider provider, string appName, int numScreens)
     {
         var controlFactory = provider.GetRequiredService<IControlFactory>();
 
-        var label1 = controlFactory.Create("Label1", template: "Label",
-            properties: new()
-            {
+        var app = controlFactory.CreateApp(appName);
+
+        for (var i = 0; i < numScreens; i++)
+        {
+            var label1 = controlFactory.Create("Label1", template: "Label",
+                properties: new()
+                {
                 { "Align", "Align.Center" },
                 { "BorderColor", "RGBA(0, 18, 107, 1)" },
                 { "Color", "RGBA(0, 0, 0, 1)" },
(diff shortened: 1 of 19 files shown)


samples/Test.AppWriter/ExampleAppGenerator.cs
```diff
@@ -0,0 +1,121 @@
...
+    public static List<(string, string)> ParseControlsInfo(string[] controlsinfo)
+    {
+        var tuples = new List<(string, string)>();
+
+        for (var i = 0; i < controlsinfo.Length; i += 2)
+        {
+            tuples.Add((controlsinfo[i], controlsinfo[i + 1]));
+        }
+
+        return tuples;
```
tehcrashxor, line 25: Style opinion: LINQ is often more concise and readable.  
(Readability of course being subjective, but I find it simpler ¯\\\_(ツ)_/¯ )

```suggestion
        return controlsinfo
            .Chunk(2)
            .Select(pair => (pair[0], pair[1]))
            .ToList();
```

samples/Test.AppWriter/Program.cs
```diff
@@ -1,16 +1,23 @@
...
 
 namespace Test.AppWriter;
 
 internal class Program
 {
+    //    private struct ControlsInfo
+    //    {
+    //        public int numScreens;
+    //        public string[] controls;
+    //    };
```
tehcrashxor, line 19: Commented out code should generally be removed, as its purpose is not clear to future maintainers (often including oneself several months down the line).

If there is a reason to keep it, such as a feature temporarily disabled or as an example or reminder, it's helpful to also put an explanatory comment above it.

samples/Test.AppWriter/Program.cs
```diff
@@ -20,99 +27,180 @@ private static IServiceProvider ConfigureServiceProvider()
...
+                    }
+                    return number;
+                }
+                else
+                {
+                    result.ErrorMessage = "Invalid value, unable to convert to integer.";
+                    return 0; // Ignored.
+                }
+            }
+            );
```
tehcrashxor: For simple types like `int`, you should be able to rely on System.CommandLine to do the parsing for you.   

Validation such as the negative number check can be added in a `.AddValidator(...)` call to separate Parsing and Validation logic.

Not that this is a big deal for small example apps, but thought you might want to be aware of the options.

samples/Test.AppWriter/Program.cs
```diff
@@ -20,99 +27,180 @@ private static IServiceProvider ConfigureServiceProvider()
...
+            isDefault: true,
+            parseArgument: result =>
+            {
+                if (result.Tokens.Count == 0)
+                {
+                    result.ErrorMessage = "Must provide a file path.";
+                    return null;
+                    //Console.WriteLine("Using default out path: ");
+                    //return new FileInfo("appname.msapp");
+                }
```
tehcrashxor, line 130: If you set the `IsDefault` property on the `Option<>`, it should handle this for you.

tehcrashxor (approved):

samples/Test.AppWriter/Program.cs
```diff
@@ -20,99 +27,180 @@ private static IServiceProvider ConfigureServiceProvider()
         return serviceProvider;
     }
 
-    // This produces a simple example app including the specified number of screens and a few generic controls
-    // This is intended for testing purposes only
-    private static App GetExampleApp(IServiceProvider provider, string appName)
+    //private static string? ValidateFilePath()
```
petrochuk, line 30: +1 to @tehcrashxor 's feedback. Commented out code should have a reason to be left here. If there is no good reason then just remove it because anyone can always go back in git history .

samples/Test.AppWriter/ExampleAppGenerator.cs
cwduncan: The structure in which this specific data will be gathered from the commandline is still a work in progress, but stylistically I have no complaints about using the suggested format.

tehcrashxor (approved):

PR #579, opened 2024-03-07
ZIndex handling - Controls should be serialized in ZIndex order, and the property itself should be ommited from the YAML

tehcrashxor:
When serializing, child controls will be sorted by their ZIndex property and the property itself will be omitted from the yaml output.  When deserializing, the ZIndex property will be added to each control based on the order found in the yaml array.

### DIFF
--- a/src/Persistence/Collections/ControlPropertiesCollection.cs
+++ b/src/Persistence/Collections/ControlPropertiesCollection.cs
@@ -79,6 +79,14 @@ public class ControlPropertiesCollection : IReadOnlyDictionary<string, ControlPr
         _properties.Add(keyValue.Item1, new ControlPropertyValue(keyValue.Item2));
     }
 
+    internal void Set(string key, ControlPropertyValue value)
+    {
+        if (string.IsNullOrWhiteSpace(key))
+            throw new ArgumentNullException(nameof(key));
+
+        _properties[key] = value;
+    }
+
     public void Remove(string key)
     {
         _properties.Remove(key);
--- a/src/Persistence/Models/Control.cs
+++ b/src/Persistence/Models/Control.cs
@@ -2,16 +2,21 @@
 // Licensed under the MIT License.
 
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using Microsoft.PowerPlatform.PowerApps.Persistence.Collections;
 using Microsoft.PowerPlatform.PowerApps.Persistence.Templates;
 using Microsoft.PowerPlatform.PowerApps.Persistence.Yaml;
 using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.Callbacks;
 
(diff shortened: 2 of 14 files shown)


src/Persistence/Models/Control.cs
```diff
@@ -56,11 +61,55 @@ public Control(string name, ControlTemplate template)
...
+        if (Children == null)
+            return;
+
+        _children = Children
+            .OrderByDescending(getZIndex)
+            .Select(removeZIndexProperty)
+            .ToArray();
+
+        static int getZIndex(Control child) =>
+            child.Properties.TryGetValue("ZIndex", out var prop) && int.TryParse(prop.Value, out var zIndex)
```
petrochuk, line 105: Can "ZIndex" be const string somewhere?

tehcrashxor: yep, done

petrochuk (approved):

src/Persistence/Collections/ControlPropertiesCollection.cs
```diff
@@ -79,6 +79,14 @@ public void Add(Tuple<string, string> keyValue)
         _properties.Add(keyValue.Item1, new ControlPropertyValue(keyValue.Item2));
     }
 
+    internal void Set(string key, ControlPropertyValue value)
```
mizrael, line 82: what if the key is not present? I'd suggest to remove this method and handle the logic via the `[]` operator instead

src/Persistence/Models/Control.cs
```diff
@@ -56,11 +61,55 @@ public Control(string name, ControlTemplate template)
...
+    public Control[] Children { get => _children; init => _children = value; }
 
     [YamlIgnore]
     public ControlEditorState? EditorState { get; set; }
 
     [YamlIgnore]
     public required ControlTemplate Template { get; init; }
+
+    [OnDeserialized]
+    internal void PostDeserialize()
```
mizrael, line 74: nit: use `Before` and `After`

tehcrashxor: done in https://github.com/microsoft/PowerApps-Tooling/pull/580

src/Persistence/Collections/ControlPropertiesCollection.cs
tehcrashxor: It doesn't matter if the key is not present - it is either added or overwritten

mizrael: true, but now we have even another way of `Add`-ing a property. "Set" is a pretty generic name and could be interpreted as an upsert operation. At this point I'd suggest to update the class interface and rename all the `Add` to `Set` to avoid confusion

tehcrashxor: > could be interpreted as an upsert operation

Correct, that's what this is.  The deserialization needs to add the ZIndex, and overwrite any existing ZIndex already there.

PR #581, opened 2024-03-08
Added support for nested templates

petrochuk:


src/Persistence.Tests/Yaml/DeserializerValidTests.cs
```diff
@@ -39,7 +39,7 @@ public class DeserializerValidTests : TestBase
         sut.Should().NotBeNull().And.BeOfType<Screen>();
         sut.Name.Should().Be("Screen1");
         sut.TemplateId.Should().Be("http://microsoft.com/appmagic/screen");
-        sut.Children.Should().NotBeNull().And.BeEmpty();
+        sut.Children.Should().BeNull();
```
mizrael, line 42: please avoid having null collections, it leads to nullref exceptions

src/Persistence/Models/Control.cs
```diff
@@ -112,4 +113,31 @@ static Control removeZIndexProperty(Control child)
...
+            for (var i = 0; i < _children.Count; i++)
+            {
+                if (_children[i].TemplateId != nestedTemplate.Id)
+                    continue;
+
+                foreach (var childTemplateProperty in _children[i].Properties)
+                {
+                    Properties.Add(childTemplateProperty.Key, childTemplateProperty.Value);
+                }
+                _children.RemoveAt(i);
```
mizrael, line 136: why are we mutating the state?

src/Persistence/Models/Control.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.PowerPlatform.PowerApps.Persistence.Models;
 public abstract record Control
 {
     [SuppressMessage("Style", "IDE0032:Use auto property", Justification = "We need both 'public init' and 'private set', which cannot be accomplished by auto property")]
-    private Control[] _children = Array.Empty<Control>();
+    private IList<Control>? _children;
```
mizrael, line 18: why?

src/Persistence/Templates/ControlFactory.cs
```diff
@@ -15,21 +15,21 @@ public ControlFactory(IControlTemplateStore controlTemplateStore)
...
-    public Control Create(string name, string template, ControlPropertiesCollection? properties = null, Control[]? children = null)
+    public Control Create(string name, string template, ControlPropertiesCollection? properties = null, IList<Control>? children = null)
     {
         if (_controlTemplateStore.TryGetTemplateByName(template, out var controlTemplate))
         {
             return new BuiltInControl(name, controlTemplate)
             {
                 Properties = properties ?? new(),
-                Children = children ?? Array.Empty<Control>()
+                Children = children
```
mizrael, line 25: please add a fallback when null

src/Persistence/Templates/ControlFactory.cs
```diff
@@ -60,12 +63,12 @@ public App CreateApp(string name, ControlPropertiesCollection? properties = null
...
     }
 
-    public Screen CreateScreen(string name, ControlPropertiesCollection? properties = null, Control[]? children = null)
+    public Screen CreateScreen(string name, ControlPropertiesCollection? properties = null, IList<Control>? children = null)
     {
         return new Screen(name, _controlTemplateStore)
         {
             Properties = properties ?? new(),
-            Children = children ?? Array.Empty<Control>()
+            Children = children
```
mizrael, line 71: please add fallback

src/Persistence/Templates/ControlTemplate.cs
```diff
@@ -54,4 +54,6 @@ public string DisplayName
     }
 
     public bool HasDisplayName => !string.IsNullOrWhiteSpace(_displayName);
+
+    public IList<ControlTemplate>? NestedTemplates { get; init; }
```
mizrael, line 58: please use an array instead

mizrael (changes requested):

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
petrochuk: we use nullable which make it very clear what can be null

src/Persistence/Models/Control.cs
petrochuk: because it needs to be modified

src/Persistence/Templates/ControlFactory.cs
petrochuk: why we need fallback here and to what?

src/Persistence/Templates/ControlFactory.cs
petrochuk: it can be null. Fallback from null to null?

src/Persistence/Templates/ControlTemplate.cs
petrochuk: List can be modified at run time

src/Persistence/Models/Control.cs
petrochuk: this is the requirement

src/Persistence/Templates/ControlFactory.cs
mizrael: that's the point, it should be empty collection

src/Persistence/Templates/ControlFactory.cs
mizrael: `Enumerable.Empty`

petrochuk: Some controls should never have children and having "null" value signifies that fact. I am not sure why we need to have empty list in these cases

src/Persistence/Templates/ControlFactory.cs
petrochuk: Added "This collection can be null in cases where the control does not support children" to the property

src/Persistence/Templates/ControlFactory.cs
mizrael: true. But from the code perspective, having the guarantee that the collection is always initialized (even when empty) saves from the headache of doing null-checks each time. It would lead to far less bugs.

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
petrochuk: If someone tries to get children on control which should never have children, nullref exception is appropriate in my opinion.

src/Persistence/Templates/ControlFactory.cs
petrochuk: having empty collection will lead to different class of bugs when user can safely enumerate children when there should be no children at all.

mizrael: this is why we should expose this as array or readonly collection and add a specific `AddChildren` method. The collection itself would be always instantiated and the Add method would check if the control can have children or not

petrochuk: I am not talking about adding children. I am talking about enumerating children when there should be no children at all. There should be an exception in this case

mizrael: sorry, but I disagree with that. We should throw when adding a child to a control that _can't_ have ones, not when asking "how many children do you have". As rule of thumb, queries should never throw, while commands instead can fail (for example when input validation goes wrong)

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
```diff
@@ -267,4 +270,29 @@ public void Deserialize_Strings()
...
+        screen.Should().NotBeNull();
+        if (screen.Children == null)
+            throw new ArgumentNullException(nameof(screen.Children));
+        screen.Children.Should().NotBeNull().And.HaveCount(1);
+        var gallery = screen.Children[0];
+        gallery.Should().NotBeNull().And.BeOfType<BuiltInControl>();
+        gallery.Template.Name.Should().Be("Gallery");
+        if (gallery.Children == null)
+            throw new ArgumentNullException(nameof(gallery.Children));
+        gallery.Children.Should().NotBeNull().And.HaveCount(2);
```
tehcrashxor, line 295: The if -> throw is redundant with the `.NotBeNull()` FluentAssertion right below it
```suggestion
        gallery.Children.Should().NotBeNull().And.HaveCount(2);
```

src/Persistence/Templates/ControlFactory.cs
tehcrashxor: Perhaps we should move `Children` up into a higher class?  Controls which may have children inherit from that, controls which don't instead inherit from the base?

Label -> Control
Gallery -> ControlWithChildren -> Control

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
petrochuk: Done

src/Persistence/Templates/ControlFactory.cs
petrochuk: Might be idea for separate PR

tehcrashxor (approved):

PR #591, opened 2024-03-19
Moved AppGenerator to shared library and refactored Generator code

cwduncan:

## Problem

## Solution

## Changes

- Moved App Generator code to shared library for access from other sample app and separation of responsibility
- Refactored App Generator code to include factory pattern for different generator implementations

## Validation

### DIFF
--- a/samples/MSAppGenerator/ExampleGenerator.cs
+++ b/samples/MSAppGenerator/ExampleGenerator.cs
@@ -6,6 +6,9 @@ using Microsoft.PowerPlatform.PowerApps.Persistence.Models;
 
 namespace MSAppGenerator;
 
+/// <summary>
+/// Generates an MSApp from a set of parameters
+/// </summary>
 public class ExampleGenerator : IAppGenerator
 {
     private readonly IControlFactory _controlFactory;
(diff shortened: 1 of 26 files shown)


petrochuk (approved):

petrochuk (approved):

samples/MSAppGenerator/ExampleGenerator.cs
```diff
@@ -1,25 +1,34 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.PowerPlatform.PowerApps.Persistence.Templates;
```
swapnilbhalgat, line 4: nit: I know it was already there, but since you're touching the file you can sort the imports. VS context menu has 'Remove and sort usings' option.

swapnilbhalgat (approved):

samples/MSAppGenerator/ExampleGenerator.cs
cwduncan: I've updated the files in MSAppGenerator to correct this, thanks for the hint on the builtin formatter

PR #592, opened 2024-03-19
added Custom Properties on Components

mizrael:
added support for Custom Properties on Components. This is a preliminary PR, YAML output is likely to change.


src/Persistence/Models/Component.cs
```diff
@@ -24,4 +24,7 @@ public Component(string name, string variant, IControlTemplateStore controlTempl
         Variant = variant;
         Template = controlTemplateStore.GetByName(BuiltInTemplates.Component);
     }
+
+    [YamlMember(Order = 4)]
```
petrochuk, line 28: It might be useful to set it to 100 to create room for more properties in base class

src/Persistence/Models/DataType.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.Models;
+
+#pragma warning disable CA1720 // Identifier contains type name
+public enum DataType
```
petrochuk, line 7: Can this get out of sync with server side?

mizrael: yes, I was thinking the same. We could use `string` instead, what do you think

petrochuk: string is safer I think

petrochuk (approved):

PR #600, opened 2024-03-24
Sorted by category and name

petrochuk:


swapnilbhalgat (approved):

mizrael:
That's a good addition, thank you. 
Just to understand the rationale: why not augmenting `ControlPropertyValue` instead? Also, are we going to delete `ControlPropertyValue` now?

petrochuk:
ControlPropertyValue just got renamed to ControlProperty

petrochuk:
@mizrael oops. my bad. I forgot to actually delete ControlPropertyValue. The original intent was to add "Name" and rename ControlPropertyValue to ControlProperty because it is no longer just value but name/value pair

PR #605, opened 2024-03-27
Improved error reporting for invalid properties

petrochuk:
Expected error message:
Property 'NoValidControl' not found on type 'Microsoft.PowerPlatform.PowerApps.Persistence.Models.Control'.

### DIFF
--- a/src/Persistence.Tests/_TestData/InvalidYaml/screen-with-not-valid-control.pa.yaml
+++ b/src/Persistence.Tests/_TestData/InvalidYaml/screen-with-not-valid-control.pa.yaml
@@ -1,0 +1,11 @@
+Screens:
+- Screen: 
+  Name: Screen1
+  Properties:
+    LoadingSpinnerColor: =RGBA(56, 96, 178, 1)
+  Children:
+  - NoValidControl: 
+    Name: Foo
+    Properties:
+      BorderColor: =RGBA(0, 18, 107, 1)
+      BorderThickness: =2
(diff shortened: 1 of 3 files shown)


src/Persistence.Tests/_TestData/InvalidYaml/screen-with-not-valid-control.pa.yaml
```diff
@@ -0,0 +1,11 @@
+Screens:
+- Screen: 
+  Name: Screen1
+  Properties:
+    LoadingSpinnerColor: =RGBA(56, 96, 178, 1)
+  Children:
+  - NoValidControl:
```
petrochuk, line 7: @lesaltzm FYI

mizrael:
I don't follow: shouldn't we be throwing when there's no template found?

petrochuk:
> I don't follow: shouldn't we be throwing when there's no template found?

We don't know if this is template or unexpected property. We will be a little more strict for controls

mizrael:
> > I don't follow: shouldn't we be throwing when there's no template found?
> 
> We don't know if this is template or unexpected property. We will be a little more strict for controls

why? if it's under the `Children` node it's definitely a Control. Maybe I am missing something ?

petrochuk:
> > > I don't follow: shouldn't we be throwing when there's no template found?
> > 
> > 
> > We don't know if this is template or unexpected property. We will be a little more strict for controls
> 
> why? if it's under the `Children` node it's definitely a Control. Maybe I am missing something ?

Here is a little better example. What should we do in this case?

``` yaml
Screens:
- Screen: 
  Name: Screen1
  Properties:
    LoadingSpinnerColor: =RGBA(56, 96, 178, 1)
  Children:
  - NoValidControl:
    StillNotValid:
    AgianNotValid:
    NotValid:
    Name: Foo
    MyName: Bar
    Properties:
      BorderColor: =RGBA(0, 18, 107, 1)
      BorderThickness: =2

```

mizrael:
took this offline. So this change is actually throwing when nodes are not recognized as valid templates.

Thanks!

mizrael (approved):

PR #607, opened 2024-03-27
Yaml Serializer cleanup and tests

tehcrashxor:
Renaming from PR feedback on #606 post merge

### DIFF
--- a/src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
+++ b/src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
@@ -52,7 +52,7 @@ public class DeserializerInvalidTests
     {
         // Arrange
         var deserializer = TestBase.ServiceProvider.GetRequiredService<IYamlSerializationFactory>().CreateDeserializer();
-        using var yamlStream = File.OpenRead("_TestData\\ValidYaml\\Screen\\with-name.pa.yaml");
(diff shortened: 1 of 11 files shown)


src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
```diff
@@ -45,4 +46,35 @@ public void Deserialize_ShouldFailWhenYamlIsInvalid()
         failedFiles.Should().BeGreaterThan(0);
         failedFiles.Should().Be(files.Length, "all files should fail");
     }
+
+    [TestMethod]
+    public void Deserialize_ShouldFailWhenExpectingDifferentType()
+    {
+        // Arrange
+        var deserializer = TestBase.ServiceProvider.GetRequiredService<IYamlSerializationFactory>().CreateDeserializer();
+        using var yamlStream = File.OpenRead("_TestData\\ValidYaml\\Screen\\with-name.pa.yaml");
```
petrochuk, line 55: nit: use `@` for path names

petrochuk (approved):

src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
tehcrashxor: Actually had to swap to `/` like the other tests, as that works for both the Windows and Linux builds

PR #609, opened 2024-03-29
Added Group control

petrochuk:


src/Persistence.Tests/_TestData/ValidYaml/Group/with-two-children.pa.yaml
```diff
@@ -0,0 +1,7 @@
+Group: 
+Name: My Small Group
+Children:
+- Button:
```
swapnilbhalgat, line 4: Is this correct identation for children items? Screenshoto of an example from new schema.
image

swapnilbhalgat (approved):

src/Persistence.Tests/_TestData/ValidYaml/Group/with-two-children.pa.yaml
petrochuk: it is correct as of now, but we will change all places at once if/when we change schema

PR #616, opened 2024-04-03
component definition custom properties as dictionary

mizrael:
Updated the implementation of custom properties on component definitions to use a dictionary instead of an array.
This allows the serializer to output YAML in this form:

```
Component: 
Name: Component1
CustomProperties:
  MyFuncProp1:
    Direction: Input
    Parameters:
      param1:
        DataType: String
        IsRequired: true
```

instead of
```
Component: 
Name: Component1
CustomProperties:
- Name: MyFuncProp1
  Direction: Input
  PropertyType: Function
  DataType: String
  Category: Data
  IsResettable: false
  Default: lorem
  Parameters:
  - Name: param1
```


src/Persistence/Models/CustomProperty.cs
```diff
@@ -28,19 +31,37 @@ public record CustomProperty
...
-    public IList<CustomPropertyParameter> Parameters { get; init; } = new List<CustomPropertyParameter>();
+    public CustomPropertyParametersCollection Parameters { get; init; } = new();
 
     public enum PropertyDirection
     {
         Input,
         Output
     }
 
+    public enum PropertyCategory
```
petrochuk: why not use https://github.com/microsoft/PowerApps-Tooling/blob/master/src/Persistence/Models/PropertyCategory.cs?

petrochuk (approved):

src/Persistence/Models/CustomProperty.cs
mizrael: I removed that, not sure why it kept showing up :)

PR #618, opened 2024-04-03
Yaml Schema v2.2 OM - initial implementation

joem-msft:
## Problem

Simple implementation of Yaml Schema v2.2 OM (as defined in XR and schema file already in master).

## Solution

Added a new OM and namespace for representing just the Yaml schema for *.pa.yaml files.
Added unit tests and also verify the Example pa.yaml files written according to the schema.

NOTE: This OM is an initial implementation to enable the start of writing the transformation from the existing OM (aka v2.1) to v2.2.

### DIFF
--- a/src/Persistence/PaYaml/Models/PowerFx/PFxDataType.cs
+++ b/src/Persistence/PaYaml/Models/PowerFx/PFxDataType.cs
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.PowerFx;
+
+public enum PFxDataType
+{
+    Text,
+    Number,
+    Boolean,
+    DateAndTime,
+    Screen,
+    Record,
+    Table,
+    Image,
+    VideoOrAudio,
+    Color,
+    Currency,
+}
--- a/src/Persistence/PaYaml/Models/SchemaV2_2/ComponentDefinition.cs
+++ b/src/Persistence/PaYaml/Models/SchemaV2_2/ComponentDefinition.cs
@@ -1,0 +1,56 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.PowerFx;
+using YamlDotNet.Serialization;
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.SchemaV2_2;
+
+public enum ComponentPropertyKind
+{
+    Input,
+    Output,
+    InputFunction,
+    OutputFunction,
+    Event,
+    Action,
+}
+
+public class ComponentDefinition
+{
+    public string? Description { get; init; }
+
+    public bool AccessAppScope { get; init; }
+
+    public Dictionary<string, ComponentCustomPropertyUnion>? CustomProperties { get; init; }
+
+    public SortedDictionary<string, PFxExpressionYaml>? Properties { get; init; }
+
+    public List<Dictionary<string, ControlInstance>>? Children { get; init; }
+}
+
+public abstract class ComponentCustomPropertyBase
+{
+    [YamlMember(DefaultValuesHandling = DefaultValuesHandling.Preserve)]
+    public required ComponentPropertyKind PropertyKind { get; init; }
+
+    public string? DisplayName { get; init; }
+
+    public string? Description { get; init; }
+}
+
+/// <summary>
+/// Represents the union of all possible properties available on any custom property.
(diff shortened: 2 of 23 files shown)


src/Persistence/PaYaml/Models/PowerFx/PFxDataType.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.PowerFx;
+
+public enum PFxDataType
```
petrochuk, line 6: We are linking with 
    <PackageReference Include="Microsoft.PowerFx.Core" Version="1.2.0" />

do we need to recreate Power Fx enums?

src/Persistence/PaYaml/Models/SchemaV2_2/ComponentDefinition.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.PowerFx;
+using YamlDotNet.Serialization;
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.SchemaV2_2;
+
+public enum ComponentPropertyKind
+{
+    Input,
```
petrochuk, line 11: I think there is enum for that already unless you want this in separate namespace

petrochuk (approved):

src/Persistence/PaYaml/Serialization/PFxExpressionYamlConverter.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        // If the script contains a substring that would cause extra yaml escaping, force it to a literal block
+        // The goal here for PFx yaml is that we avoid using yaml string escaping whenever possible.
+        var forceLiteralBlock =
+            // These char sequences are special to YAML parsers, so must be escaped or forced to literal block
+            expression.InvariantScript.Contains(" #")
+            || expression.InvariantScript.Contains(": ")
+            // Force multi-line scripts to be literal blocks
+            || expression.InvariantScript.Contains('\n')
+            || expression.InvariantScript.Contains('\r');
```
tehcrashxor, line 60:
```suggestion
            // Force multi-line scripts to be literal blocks
            || expression.InvariantScript.Contains('\n')
            || expression.InvariantScript.Contains('\r')
            || expression.InvariantScript.Contains('\u0085')  // Unicode next line
            || expression.InvariantScript.Contains('\u2028')  // Unicode line separator
            || expression.InvariantScript.Contains('\u2029'); // Unicord paragraph separator
```

tehcrashxor (approved):

src/Persistence/PaYaml/Models/PowerFx/PFxDataType.cs
joem-msft: these models are owned by this assembly model. It's a Power Apps owned model and implementation.
It's a dangerous thing to set serialization OM to a dependancy, as changes in the dependency may break logic here.
For this, it MAY be possible to create a shared implementation, but these OMs are designed for YamlDotNet usage, and PFx doesn't use that.

joem-msft: not going to change this in this PR.

src/Persistence/PaYaml/Models/SchemaV2_2/ComponentDefinition.cs
joem-msft: this is an enum specifically mapping to the V2.2 schema.
Serialization OM != Runtime OM. ;)

joem-msft: won't fix

src/Persistence/PaYaml/Serialization/PFxExpressionYamlConverter.cs
joem-msft: turns out that's actually problematic.
Since the original code doesn't handle these special Unicode chars, I'm going to leave it out for future work.

joem-msft: won't fix.

tehcrashxor: > Since the original code doesn't handle these special Unicode chars

Incorrect, it was handled in [ControlPropertyConverter.cs](https://github.com/microsoft/PowerApps-Tooling/blob/6e96cccd53d30ec52a034b076fa414a3769d95c7/src/Persistence/Yaml/ControlPropertyConverter.cs#L13)

PR #619, opened 2024-04-03
fix local build issues

joem-msft:
## Problem

- In some developer configurations, nuget is not able to restore packages due to custom user scoped nuget configuration.
- When developing in VS IDE, builds often fail due to style rules failing. This hinders speed of development when trying to just get some things done.

## Solution

- Added nuget.config to the root of the repo which states to use nuget.org feed.
- IDE#### rules no longer error out, but are set as warnings.
- modified the msbuild properties which were forcing all analysis and style rules to cause errors, thus build failures. These now only get set when NOT in a VS IDE.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,7 +29,8 @@ indent_size = 4
 end_of_line = crlf
 trim_trailing_whitespace = true
 
-dotnet_analyzer_diagnostic.category-Style.severity = error
+# Do not ever set to error, or these cause build failures during development.
+dotnet_analyzer_diagnostic.category-Style.severity = warning
 
 # Organize usings
 dotnet_sort_system_directives_first = true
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -29,35 +29,41 @@ indent_size = 4
...
 # Organize usings
 dotnet_sort_system_directives_first = true
 dotnet_separate_import_directive_groups = false
 
 file_header_template = Copyright (c) Microsoft Corporation.\nLicensed under the MIT License.
 
-dotnet_diagnostic.IDE0003.severity = error    # this and Me preferences
-dotnet_diagnostic.IDE0007.severity = error    # Use 'var' instead of explicit type
-dotnet_diagnostic.IDE0005.severity = error    # Remove unnecessary using directives
+dotnet_diagnostic.IDE0003.severity = warning    # this and Me preferences
```
petrochuk, line 41: we don't want "this.", "Me" why change it?

.editorconfig
```diff
@@ -29,35 +29,41 @@ indent_size = 4
...
+dotnet_diagnostic.IDE0007.severity = warning    # Use 'var' instead of explicit type
+dotnet_diagnostic.IDE0005.severity = warning    # Remove unnecessary using directives
```
petrochuk, line 43: this one keeps usings clean

tehcrashxor (approved):

.editorconfig
joem-msft: this isn't removed. it's moved so all IDE* rules are in one place.
They are also made as a warning to not block development progress.

.editorconfig
joem-msft: this isn't removed. it's moved so all IDE* rules are in one place.
They are also made as a 'warning' to not block development progress.

PR #620, opened 2024-04-05
Yaml Schema OM v2.2 utilizes NamedObject, NamedObjectMapping and NamedObjectSequence

joem-msft:
## Problem

The schema for the XR yaml (v2.2) has a pattern of representing named objects such that the name is in the breadcrumb. Sometimes there are sets of named objects with no explicit ordering (other than normalized sorting by name for deterministic output) which get rendered in YAML as a mapping. When these sets of objects are supposed to have an imposed order, then they are rendered in YAML as a sequence of mappings, where each mapping has a single item (the name : value).

## Solution

Since this pattern is used all over in the schema, for control properties (PropertyName: PFx expression), Children controls, component definition custom properties, etc. We implemented the `NamedObject<TValue>` type to represent an object with a name. These instances are intended to be kept together with the object instance.

The `NamedObjectMapping<TValue>` is a collection of `NamedObject<TValue>` instances, but is represented in Yaml as a mapping. The order of the items in this collection is not maintained, nor is significant. But, in order to maintain consistent normalized yaml output, the items are sorted by their names.

The `NamedObjectSequence<TValue>` is also a collection of `NamedObject<TValue>` instances, but is represented in Yaml as a sequence. The sequence encodes the order of each object being significant. e.g. child controls have an ordering, which a ZIndex can be computed.

## Changes

NO changes in this PR are breaking for the DocumentServer as these new OMs are not currently used by anything other than unit tests.

- Updated new PaYaml models to utilize the new NamedObjectMapping/NamedObjectSequence data structures to provide a consistent API surface.
- Moved some of the new OM classes a bit, and moved anything related to serialization into the `PaYaml.Serialization` namespace.
- Implemented YamlDotNet converter logic to enable serializing of the different types.
- Implemented extensions to FluentAssertions to support working with NamedObject collections.

## Validation

- Updated existing tests to use the new OMs
- Created new tests to target some of the more complex data structures.


.editorconfig
```diff
@@ -545,7 +545,7 @@ dotnet_diagnostic.CA1720.severity = error
 dotnet_diagnostic.CA1810.severity = error
 dotnet_diagnostic.CA1821.severity = error
 dotnet_diagnostic.CA1822.severity = error
-dotnet_diagnostic.CA1823.severity = error
```
petrochuk, line 548: Can you leave this as error?

joem-msft: Setting to warning is appropriate for development experience. During any non-ide build, warnings are turned into errors. (see msbuild properties `TreatWarningsAsErrors` and `CodeAnalysisTreatWarningsAsErrors`) This is a recommended pattern that speeds up local development.

joem-msft: won't fix

src/Persistence/PaYaml/Models/SchemaV2_2/ModelsExtensions.cs
```diff
@@ -14,11 +14,8 @@ public static int GetDescendantsCount(this ScreenInstance screen)
...
 
-            foreach (var namedDict in screen.Children)
+            foreach (var namedControl in screen.Children)
             {
-                Debug.Assert(namedDict.Count == 1, "Each child of a screen should represent a dictionary with a single item.");
-                var namedControl = namedDict.Single();
-
                 count += namedControl.Value.GetDescendantsCount();
             }
         }
```
tehcrashxor, line 21: Could remove the `if-foreach` pattern with:
```csharp
return screen.Children?.Count > 0
    ? screen.Children.Count + screen.Children.Sum(namedControl => namedControl.Value.GetDescendantsCount())
    : 0;
```

tehcrashxor: ditto below

src/Persistence.Tests/PaYaml/Serialization/PaYamlSerializerTests.cs
```diff
@@ -95,10 +95,18 @@ public void DeserializeExamplePaYamlSingleFileApp()
...
 
         // Check random spot in the document
-        paFileRoot.App.Properties.Should().ContainKey("Theme");
-        var screen = paFileRoot.Screens.Should().ContainKey("Screen1").WhoseValue;
+        var dict = new Dictionary<string, string>
+        {
+            { "Theme", "Light" },
+            { "Title", "Single-File App" }
+        };
+        dict.Should().ContainKey("Theme");
```
tehcrashxor, line 103: Debugging code to remove?

tehcrashxor (approved):

src/Persistence/PaYaml/Models/SchemaV2_2/ModelsExtensions.cs
joem-msft: woah! Copilot totally just recommended that exact expression! magic

PR #622, opened 2024-04-05
Reverse ZIndex ordering for children of group containers

lesaltzm:
Children of group containers are ordered in the tree view in reverse of zindex order. This update changes the serialize/deserialize logic so that the serialized order matches the tree view order.

### DIFF
--- a/src/Persistence/Models/Control.cs
+++ b/src/Persistence/Models/Control.cs
@@ -101,9 +101,18 @@ public abstract record Control
         if (this is App)
             return; // Apps do not place ZIndex on their Host child
 
+
+        var zindexCalc = (int i) => Children.Count - i;
+        // For group containers, ZIndex is the actual order of what is shown in the tree view, not descending.
+        // Handle that here to ensure we match what is expected
+        if (Template.Name == "GroupContainer")
+        {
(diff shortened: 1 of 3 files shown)


src/Persistence/Models/Control.cs
```diff
@@ -113,8 +122,18 @@ internal void BeforeSerialize()
     {
         HideNestedTemplates();
 
+        Comparison<Control> zIndexComparison = (Control c1, Control c2) => c2.ZIndex.CompareTo(c1.ZIndex);
+
+        // For group containers, ZIndex is the reverse order of what is shown in the tree view
+        // Handle that here to ensure we match what is expected
+        if (Template.Name == "GroupContainer")
```
lesaltzm, line 129: Should this constant live somewhere?

mizrael: https://github.com/microsoft/PowerApps-Tooling/blob/master/src/Persistence/Templates/BuiltInTemplates.cs maybe?

mizrael (changes requested):

mizrael (approved):

tehcrashxor (approved):

PR #626, opened 2024-04-09
updated Component Definitions

mizrael:
- added `Description` and `AccessAppScope` to Component Definitions
- removed `ToolTip` property from Custom Properties


src/Persistence/Yaml/ComponentConverter.cs
```diff
@@ -34,6 +34,18 @@ void IYamlTypeConverter.WriteYaml(IEmitter emitter, object? value, Type type)
         var component = ((Component)value).BeforeSerialize<Component>();
         WriteYamlInternal(emitter, component, type);
 
+        if (!string.IsNullOrWhiteSpace(component.Description))
```
petrochuk, line 37: there is Emit extension method that does that YamlExtensions.cs

petrochuk (approved):

PR #630, opened 2024-04-10
Invert Zindex order for children of group controls

lesaltzm:
It looks like #623 changed how ZIndices are applied to controls, so #622 was no longer in effect. This brings back the reordering of zindex for children of group containers.


src/Persistence/Models/Control.cs
```diff
@@ -94,6 +94,8 @@ public int ZIndex
     [OnDeserialized]
     internal virtual void AfterDeserialize()
     {
+        // Is this dead code now? Looks like it's not being hit anymore after the schema changes?
```
lesaltzm, line 97: Will clean up this comment, but wanted to check whether this was dead code or not?

mizrael: @anpetroc can you chime in? I was trying to debug this method this morning, but it doesn't get called (at least not when running the tests). If that's the case, I'd say we can get rid of it

src/Persistence.Tests/_TestData/ValidYaml/ZIndexOrdering/With-list-of-container-controls-with-children.pa.yaml
```diff
@@ -0,0 +1,22 @@
+- GroupContainer: 
+  Name: Group1
```
petrochuk, line 2: this one is missing

`Control: Group`

src/Persistence/Models/Control.cs
petrochuk: It probably is dead code now as we are switching to some new object model hooks

petrochuk (approved):

PR #633, opened 2024-04-11
CustomProperty ordering

petrochuk:


swapnilbhalgat (approved):

src/Persistence/Models/Component.cs
```diff
@@ -39,11 +38,11 @@ public Component(string name, string variant, IControlTemplateStore controlTempl
     public bool AccessAppScope { get; set; }
 
     [YamlMember(Order = 100)]
-    public CustomPropertiesCollection CustomProperties { get; set; } = new();
+    public IList<CustomProperty> CustomProperties { get; set; } = new List<CustomProperty>();
 
     internal override void AfterCreate(Dictionary<string, object?> controlDefinition)
     {
-        if (controlDefinition.TryGetValue<CustomPropertiesCollection>(nameof(CustomProperties), out var customProperties))
+        if (controlDefinition.TryGetValue<List<CustomProperty>>(nameof(CustomProperties), out var customProperties))
```
mizrael, line 45: nit: since `CustomProperties` has a public setter, might be safer to cast this to `IList<CustomProperty>`

src/Persistence/Models/Component.cs
```diff
@@ -39,11 +38,11 @@ public Component(string name, string variant, IControlTemplateStore controlTempl
     public bool AccessAppScope { get; set; }
 
     [YamlMember(Order = 100)]
-    public CustomPropertiesCollection CustomProperties { get; set; } = new();
+    public IList<CustomProperty> CustomProperties { get; set; } = new List<CustomProperty>();
```
mizrael, line 41: are we going to delete the `CustomPropertiesCollection` now?

src/Persistence/Yaml/INamedObject.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.Yaml;
+
+public interface INamedObject
+{
+    string Name { get; set; }
```
mizrael, line 8: nit: if possible, I'd remove the setter from the interface (unless strictly necessary)

src/Persistence/Yaml/NamedObjectEmitter.cs
```diff
@@ -14,38 +14,54 @@ namespace Microsoft.PowerPlatform.PowerApps.Persistence.Yaml;
 ///    Property1: Value1
 ///    Property2: Value2
 /// </summary>
-/// <typeparam name="T"></typeparam>
-internal class NamedObjectEmitter<T> : ChainedEventEmitter
-    where T : class
+internal class NamedObjectEmitter : ChainedEventEmitter
 {
-    private readonly Func<T, string?> _nodeNameProvider;
+    private Stack<bool> _isNamedObject = new Stack<bool>();
```
mizrael, line 19: what's the lifetime of `NamedObjectEmitter` ? What about multithreading?

src/Persistence/Yaml/NamedObjectTypeInspector.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly ITypeInspector _innerTypeInspector;
+
+    public NamedObjectTypeInspector(ITypeInspector innerTypeInspector)
+    {
+        _innerTypeInspector = innerTypeInspector;
+    }
+
+    public IEnumerable<IPropertyDescriptor> GetProperties(Type type, object? container)
+    {
+        var properties = _innerTypeInspector.GetProperties(type, container);
```
mizrael, line 19: not for this PR, but we should see if caching properties brings some benefits

src/Persistence/Yaml/NamedObjectsCollectionConverter.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    public IValueDeserializer? ValueDeserializer { get; set; }
+
+    public IValueSerializer? ValueSerializer { get; set; }
+
+    public bool Accepts(Type type)
+    {
+        return
+            type == typeof(List<T>) || type.IsSubclassOf(typeof(List<T>)) ||
+            type == typeof(IList<T>) || type.IsSubclassOf(typeof(IList<T>)) ||
```
mizrael, line 24: I suspect that `IsSubclassOf(typeof(IList<T>))` is not really doing much. Being `IList` an interface, it should _probably_ be `IsAssignableFrom`

src/Persistence/Yaml/NamedObjectsCollectionConverter.cs
```diff
@@ -0,0 +1,93 @@
...
+                    throw new YamlException(parser.Current.Start, parser.Current.End, $"Expected definition of {typeof(T).Name}");
+                if (parser.Current is not Scalar)
+                    throw new YamlException(parser.Current!.Start, parser.Current.End, $"Expected name of {typeof(T).Name}");
+
+                name = parser.Consume<Scalar>().Value;
+            }
+
+            using var serializerState = new SerializerState();
+            var controlObj = ValueDeserializer!.DeserializeValue(parser, typeof(T), serializerState, ValueDeserializer);
+            if (controlObj == null || controlObj is not T)
```
mizrael, line 54: nit: `controlObj is not T namedObject` avoids casting again later on

src/Persistence/Models/Component.cs
petrochuk: I think so because NamedObjectsCollectionConverter can handle any named objects collection which implement simple INamedObject interface

src/Persistence/Yaml/NamedObjectEmitter.cs
petrochuk: This emitter is not actually used at all. I kept it for future reference

src/Persistence/Yaml/NamedObjectTypeInspector.cs
petrochuk: _innerTypeInspector is actually is caching it already

src/Persistence/Yaml/INamedObject.cs
petrochuk: I need to call set during serialization. If you find a way to improve feel free to do it

src/Persistence/Models/Component.cs
petrochuk: Done

src/Persistence/Yaml/NamedObjectsCollectionConverter.cs
petrochuk: Interfaces behave differently. It works with this logic. If you know how to improve let's change it

src/Persistence/Yaml/NamedObjectsCollectionConverter.cs
petrochuk: done

mizrael (approved):

PR #637, opened 2024-04-13
Add Layout to schema and PaYaml OM

joem-msft:
- Added `Layout` to top-level properties of a control instance
- Reordered `Screens` and `ComponentDefinitions` in PaFileRoot so the order puts the dependencies first.


petrochuk (approved):

joem-msft:
I'm going to abandon this PR as it's come to my attention that the Layout property may not actually be needed. Will add it in the future as part of the work to implement the XR Schema.

PR #638, opened 2024-04-13
`WellKnownTemplateIds` and `AddPowerAppsPersistence` refactoring

joem-msft:
- `WellKnownTemplateIds` constants for id's used in the persistence assembly. Updated ServiceCollectionExtensions to use these.
- `AddPowerAppsPersistence` modified:
  - accepts additional parameter for configuring template store. This allows callers unrestricted ability to build the template store removing need to add code in product implementation for testing setup logic.
  - removed parameter `useDefaultTemplates` in preference for tests to make explicit call to utility method or to do other custom logic.
  - Renamed `AddDefaultTemplates` to `TESTING_ONLY_AddDefaultTemplates` to be clear about intention of usage. Made it public so it can be called by DocServer test code directly.
- Revert "Update `ControlTemplate` to indicate whether it's a Component or Pcf control template (#631)"
- Fixed AppGenerator to use ProjectReference to Persistence.csproj to ensure correct nuget dependency validation etc.

### DIFF
--- a/samples/Test.AppWriter/Test.AppWriter.csproj
+++ b/samples/Test.AppWriter/Test.AppWriter.csproj
@@ -21,7 +21,6 @@
     <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.0" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageReference Include="System.CommandLine.DragonFruit" Version="0.4.0-alpha.22272.1" />
-    <PackageReference Include="YamlDotNet" Version="15.1.1" />
   </ItemGroup>
 
   <ItemGroup>
@@ -30,9 +29,7 @@
 
   <ItemGroup>
     <!-- This reference will be replaced with Nuget package reference -->
-    <Reference Include="Microsoft.PowerPlatform.PowerApps.Persistence">
-      <HintPath>..\..\bin\Debug\Persistence\Microsoft.PowerPlatform.PowerApps.Persistence.dll</HintPath>
-    </Reference>
+  <ProjectReference Include="..\..\src\Persistence\Microsoft.PowerPlatform.PowerApps.Persistence.csproj" />
   </ItemGroup>
 
 </Project>
--- a/src/Persistence/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Persistence/Extensions/ServiceCollectionExtensions.cs
@@ -13,8 +13,14 @@ public static class ServiceCollectionExtensions
     /// registers the MSAPP persistence services
     /// </summary>
     /// <param name="services">the services collection instance.</param>
-    /// <param name="useDefaultTemplates">if true, registers the default templates (eg. 'text', 'button') on the templates store.</param>
-    public static void AddPowerAppsPersistence(this IServiceCollection services, bool useDefaultTemplates = false)
+    /// <param name="additionalTemplateStoreConfiguration">
+    /// When specified, performs additional configuration of the <see cref="ControlTemplateStore"/> instance.<br/>
+    /// e.g. Tests could use the following to add test templates:<br/>
+    /// <code>serviceCollection.AddPowerAppsPersistence(store => store.TESTING_ONLY_AddDefaultTemplates())</code>
+    /// </param>
+    public static void AddPowerAppsPersistence(
+        this IServiceCollection services,
+        Action<ControlTemplateStore>? additionalTemplateStoreConfiguration = null)
     {
         services.AddSingleton<IMsappArchiveFactory, MsappArchiveFactory>();
         services.AddSingleton<IYamlSerializationFactory, YamlSerializationFactory>();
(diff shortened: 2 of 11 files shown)


src/Persistence/Templates/WellKnownTemplateIds.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.Templates;
+
+public static class WellKnownTemplateIds
```
mizrael, line 6: @petrochuk  has added something similar in https://github.com/microsoft/PowerApps-Tooling/pull/636/files#diff-ac4572e52bc9a7f059eddb37ee233946299b268991310aaf4caefd7c1484ad22 . I'd wait for his PR to go in first and then adapt the changes.

src/Persistence/Templates/WellKnownTemplateIds.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public const string appinfo = "http://microsoft.com/appmagic/appinfo";
```
mizrael, line 8: also, why the casing is not consisten?

src/Persistence/Templates/WellKnownTemplateIds.cs
petrochuk: I think it makes more sense to have names and ids in one place instead of having two separate classes: one for names and one for ids

joem-msft: This class only hods IDs. the names are extra metadata which are out of scope of document template Ids.
I've remerged with master and used these constants in your 'BuiltInTemplates' class so there's still a single source of truth.

src/Persistence/Templates/WellKnownTemplateIds.cs
joem-msft: the casing is consistent with the case in the template uri. I know this isn't normal for .net naming. It's a reminder to the consumer what it's value is.

But I can make them all start with upper case, and actually, I can make them CamelCase too.

src/Persistence/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -64,15 +68,18 @@ private static void AddMinimalTemplates(ControlTemplateStore store)
...
+        store.Add(new() { Name = "commandComponent", Id = WellKnownTemplateIds.CommandComponent });
     }
 
     /// <summary>
     /// Adds some default templates which are useful for testing
     /// </summary>
     /// <param name="store"></param>
-    private static void AddDefaultTemplates(ControlTemplateStore store)
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "TEST ONLY function")]
+    public static void TESTING_ONLY_AddDefaultTemplates(this ControlTemplateStore store)
```
mizrael, line 79: please mark this as `internal`. The project is already configured to use `InternalsVisibleTo` with the tests project

mizrael (changes requested):

samples/Test.AppWriter/Test.AppWriter.csproj
```diff
@@ -30,9 +29,7 @@
 
   <ItemGroup>
     <!-- This reference will be replaced with Nuget package reference -->
-    <Reference Include="Microsoft.PowerPlatform.PowerApps.Persistence">
-      <HintPath>..\..\bin\Debug\Persistence\Microsoft.PowerPlatform.PowerApps.Persistence.dll</HintPath>
-    </Reference>
+  <ProjectReference Include="..\..\src\Persistence\Microsoft.PowerPlatform.PowerApps.Persistence.csproj" />
```
cwduncan, line 32: Clarifying question: Will this change affect the build process for the sample apps? (i.e. will Persistence be rebuilt when building the Samples solution)

src/Persistence/Extensions/ServiceCollectionExtensions.cs
joem-msft: there is a bug due to InternalsVisisbleTo not actually working due to someone disabling the wrong feature in SDK projects for this repo in order to implement versioning. We'll have to address it in the future.
[Bug 27469059](https://dev.azure.com/msazure/OneAgile/_workitems/edit/27469059): Maintainability: Repo versioning must use SDK proj properties; and reenable GenerateAssemblyGeneration

Note, this function will need to be called by the Doc Server Persistence tests too. Adding a prefix like the one here does ensure usage is clear.

samples/Test.AppWriter/Test.AppWriter.csproj
joem-msft: it will be rebuilt if it has changes, msbuild will take care of it.

joem-msft: there will be a slight general msbuild overhead, but no actual recompilation will happen.

mizrael (approved):

src/Persistence/Templates/WellKnownTemplateIds.cs
petrochuk: I think it would be useful to have ids in the same place as names

petrochuk (changes requested):

joem-msft:
abandoning this PR due to andrew's rejections

PR #640, opened 2024-04-17
Reorganize .editorconfig (no changes to rule severities)

joem-msft:
## Problem

.editorconfig rules config line orderings have caused certiain rules to get skipped.

## Solution

Changed the ordering in the editorconfig so settings are grouped and rule severities come after styling rules.
I then fixed existing code that did not comply with the rules.

## Changes

Reorganize .editorconfig (no changes to rule severities)
This commit contains no actual changes to the rules; it's only a reordering of the different types of configuration settings.
    
- Add comments to some rule codes to make them readable
- Moved all styling, naming, convention rules first. Some IDEs have been known to only support the top-most .editorconfig file to have these
- All rule severity configurations are grouped by kind:
  - .net styles
  - C# styles
  - Code-style Diagnostic Rules
  - Code Quality Diagnostic Rules
  - Style Cop Analyzer Diagnostic Rules
- Address failures that got skipped due to conflicting editorconfig ordering

## Validation

Addressed build failures and fixed unit tests.


src/Persistence/Yaml/ControlObjectFactory.cs
```diff
@@ -89,8 +89,7 @@ internal void RestoreNestedTemplates(Control control)
                 continue;
 
             var nestedControl = _controlFactory.Create(Guid.NewGuid().ToString(), nestedTemplate);
-            if (control.Children == null)
-                control.Children = new List<Control>();
```
petrochuk, line 93: It is easier to put break points when we have two lines

src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
```diff
@@ -21,31 +21,13 @@ public void Deserialize_ShouldFailWhenYamlIsInvalid(bool isControlIdentifiers)
         // Uncomment to test single file
         // var files = new string[] { @"_TestData/InvalidYaml/Screen-with-host.pa.yaml" };
 
-        var failedFiles = 0;
-        Parallel.ForEach(files, file =>
+        foreach (var filePath in files)
         {
-            using var yamlStream = File.OpenRead(file);
```
petrochuk, line 27: I am curious why you are removing parallel and making tests sequential.

src/Persistence/Yaml/ControlObjectFactory.cs
joem-msft: this was a fix based on rule errors.
VS breakpoints can be set within a line.

src/Persistence.Tests/Yaml/DeserializerInvalidTests.cs
joem-msft: parallel execution is not needed for this test. parallelize at the top layer instead. the new test code is simpler and more explicit about what's being expected than previous.

note, I only changed this here because of a couple rule failures. Turns out this test wasn't very good at describing the expectation of an exception being expected either.

joem-msft: this unit test will be updated when I implement the swallowing of all yamldotnet exceptions too

joem-msft: * I mean, parallelize at the test runner level, not within a test.
*

src/Persistence/Yaml/ControlObjectFactory.cs
petrochuk: thanks for letting me know how to set breakpoints

src/Persistence/MsApp/MsappArchive.cs
```diff
@@ -134,7 +134,7 @@ public MsappArchive(Stream stream, ZipArchiveMode mode, bool leaveOpen, Encoding
             foreach (var entry in ZipArchive.Entries)
             {
                 if (!canonicalEntries.TryAdd(NormalizePath(entry.FullName), entry))
-                    _logger?.LogInformation($"Duplicate entry found in archive: {entry.FullName}");
+                    _logger?.LogInformation("Duplicate entry found in archive: {EntryFullName}", entry.FullName);
```
mizrael, line 137: this will improve our structured logging, thank you for taking care of it!

mizrael (approved):

src/Persistence/MsApp/MsappArchive.cs
joem-msft: thank you to rules!

tehcrashxor (approved):

PR #645, opened 2024-04-18
`PersistenceException` standardization

joem-msft:
## Problem

We need to standardize on the custom exceptions that we throw and allow the DocServer to emit actionable error messages to the user if possible.

## Solution

This PR adapts our existing `PersistenceException` to require an ErrorCode and allows it to be used in more general scenarios.
We also shore up the APIs for serializing/deserializing yaml and json to wrap exceptions coming from the serialization libraries.

## Changes

- Added required `ErrorCode` property
- Added enum `PersistenceErrorCode` for consumption by DocSvr
- Made context-aware properties be explicit to the type of value and optional.
- Added `JsonPath` property
- Added `MsappEntryFullPath` property. only to be set when an entry is known.

- Updated places where exceptions are thrown.
- Added try/catches to YamlSerializer/YamlDeserializer to wrap YamlDotNet exceptions.
- Updated MsappArchive to add context and error codes.
- Updated unit tests to expect the new exception type.
- Added fluent assertion extensions to make checking error code easier.

- Fixed bug in `RoundTripWriter` dispose method which would hide original exception.
- Fixed bug in return type of `IYamlDeserializer.Deserialize` so that it's nullable. This is the correct behavior of how YamlDotNet deserializes.
  - Added unit tests showing this behavior, that it's possible for the deserialized value to be null.
  - Removed check that would throw if theyaml is an empty string. This will get deserialized as a null value. Callers should be handling null results to avoid `NullReferenceExceptions`.

## Validation

Tests updated and new ones added.


tehcrashxor (approved):

src/Persistence/Yaml/YamlSerializer.cs
```diff
@@ -19,21 +20,57 @@ public string Serialize(object graph)
...
+        }
+    }
+
+    private string SerializeCore<T>(T graph)
+    {
+        try
+        {
+            return _serializer.Serialize(graph);
+        }
+        catch (YamlException ex)
```
petrochuk, line 65: Should we catch all instead of just YamlException?

joem-msft: no. Please see: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1031

.github/workflows/CI.yml
```diff
@@ -10,6 +10,7 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.os == 'ubuntu-latest' }}
```
petrochuk, line 13: In my opinion if ubuntu build fails CI build should fail too.

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
```diff
@@ -36,7 +50,8 @@ public class DeserializerValidTests : TestBase
         var deserializer = ServiceProvider.GetRequiredService<IYamlSerializationFactory>().CreateDeserializer(new() { IsTextFirst = true });
 
         var sut = deserializer.Deserialize<Control>(yaml);
-        sut.Should().NotBeNull().And.BeOfType<Screen>();
+        sut.ShouldNotBeNull();
```
petrochuk, line 53: Is there a reason to split this into two lines? I think the test previously was doing exactly the same thing

petrochuk (approved):

src/Persistence/MsApp/MsappArchive.cs
```diff
@@ -596,6 +567,26 @@ private void CreateGitIgnore()
         writer.WriteLine("/[Aa]pp[Cc]hecker[Rr]esult.sarif");
     }
 
+    private static T DeseraializeMsappJsonFile<T>(ZipArchiveEntry entry)
```
mizrael, line 570: nit: `DeserializeMsappJsonFile`

mizrael (approved):

.github/workflows/CI.yml
joem-msft: This setting change just makes it so that when linux fails, the other jobs won't get cancelled. The job will still fail the CI build.

Linux builds continually have different implementation for code formatting errors. :( Namely formatting of comments.
This change makes it so that when it does fail, the Windows build and tests will also continue to completion. Allowing for getting more information about any failures with the same commit.

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
joem-msft: The 'ShouldNotBeNull' function is a custom extension that I created a bit ago that makes it possible to do null check validation in nullable contexts. Without this, the compiler complains about possible null dereference errors because the value returned from the previous line is nullable.

It's too bad that the FluentAssertions don't make it possible to tell the compiler about an assertion also ensuring nullable checks are done.

PR #647, opened 2024-04-20
Enable rules for IDisposable and exceptions

joem-msft:
## Problem

For code quality, maintenance, and usage this PR enables diagnostic rules related to usage of IDisposable objects and exceptions.

## Solution

Update .editorconfig rules and fixed or suppressed existing violations.

### DIFF
--- a/targets/Program.cs
+++ b/targets/Program.cs
@@ -13,7 +13,7 @@ using static SimpleExec.Command;
 
 namespace targets
 {
-    class Options
+    internal sealed class Options
     {
         [Value(0, MetaName = "target", HelpText = "build target to run; see available with: '--list", Default = "rebuild")]
         public string Target { get; set; }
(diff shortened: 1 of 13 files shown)


targets/Program.cs
```diff
@@ -11,149 +11,150 @@
...
-namespace targets
+namespace targets;
+
+internal sealed class Options
 {
-    class Options
-    {
-        [Value(0, MetaName = "target", HelpText = "build target to run; see available with: '--list", Default = "rebuild")]
-        public string Target { get; set; }
+    [Value(0, MetaName = "target", HelpText = "build target to run; see available with: '--list", Default = "rebuild")]
```
petrochuk, line 18: Why do we need to do these changes? We are not using it

joem-msft: It's a simple change to make them compliant with the new enabled rules. Instead of disabling the rules, keeping them enabled.
but I don't need to change the file scoped naming and formatting. I'll undo that so it's clearer the actual change.

tehcrashxor (approved):

PR #648, opened 2024-04-21
Added support for classic and modern controls

petrochuk:


src/Persistence.Tests/_TestData/ValidYaml-CI/BuiltInControl/with-template-name.pa.yaml
```diff
@@ -1,4 +1,4 @@
 button with template name:
-  Control: Button 
+  Classic: Button
```
mizrael, line 2: we don't have `Classic:` in the schema. Let's discuss during standup

mizrael (changes requested):
let's discuss during standup

src/Persistence.Tests/_TestData/ValidYaml-CI/BuiltInControl/with-template-name.pa.yaml
petrochuk: I don't think the goal is to have every possible property in the schema. For example, we don't have Width, Hight, X, Y, etc

petrochuk:
> let's discuss during standup

Can you clarify what changes are you requesting for this PR?

src/Persistence.Tests/_TestData/ValidYaml-CI/BuiltInControl/with-template-name.pa.yaml
joem-msft: Width, Height, X, Y are not in the schema. correct. they are extension points under the 'Properties' bag.
But this property is very much in the schema.

As to the merits of this proposal, this sort of thing was explicitly rejected in the schema discussions when we had a similar proposal for Component instances.

For context, the proposal looked something like this: 
For a Component defined in the local app with a name of `MyHeader`, a control instance would look like:
```yaml
- MyHeader_1:
    Component: MyHeader
    Properties:
      Title: ="My cool app"
      X: =0
      Y: =0
      Width: =Parent.Width
```

This was rejected by this team and architects for the reason of perceived confusion and complexity for users. Namely, that the control instance doesn't look like all other control instances.
Note, a Component instance is definitely a 1st class construct, which I believe warrants a custom implementation. Yet conceded due to overwhelming concerns.

So we kept it looking like:
```yaml
- MyHeader_1:
    Control: Component
    ComponentName: MyHeader
    Properties:
      Title: ="My cool app"
      X: =0
      Y: =0
      Width: =Parent.Width
```

Now, for classic vs modern control instances, there isn't a 1st class control type. We're only trying to resolve a unique invariant name for the control type. So I would not suggest making these controls use a new schema keyword for it just because we've got some naming conflicts.

I feel like we should have this discussion with McCall and other doc server team members in order to have a plan.

joem-msft (changes requested):

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
```diff
@@ -251,11 +249,9 @@ public void Deserialize_ShouldParseBuiltInControlFromYamlCustomControl(string te
     [DataRow(@"_TestData/ValidYaml{0}/Screen/with-template-name.pa.yaml", true, typeof(Screen), "http://microsoft.com/appmagic/screen", "Hello", 0, 0)]
     [DataRow(@"_TestData/ValidYaml{0}/Screen/with-template-name.pa.yaml", false, typeof(Screen), "http://microsoft.com/appmagic/screen", "Hello", 0, 0)]
     [DataRow(@"_TestData/ValidYaml{0}/BuiltInControl/with-template.pa.yaml", true, typeof(BuiltInControl), "http://microsoft.com/appmagic/button", "button with template", 0, 1)]
-    [DataRow(@"_TestData/ValidYaml{0}/BuiltInControl/with-template.pa.yaml", false, typeof(BuiltInControl), "http://microsoft.com/appmagic/button", "button with template", 0, 1)]
     [DataRow(@"_TestData/ValidYaml{0}/BuiltInControl/with-template-id.pa.yaml", true, typeof(BuiltInControl), "http://microsoft.com/appmagic/button", "button with template id", 0, 1)]
```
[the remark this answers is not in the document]
petrochuk: Accidental. Restored

src/Persistence.Tests/Yaml/DeserializerValidTests.cs
```diff
@@ -370,7 +366,7 @@ public void Deserialize_Strings()
         component.Description.Should().Be(expectedDescription);
         component.AccessAppScope.Should().Be(expectedAccessAppScope);
         component.Template.Should().NotBeNull();
-        component.Template!.Name.Should().Be(expectedName);
+        component.Template!.InvariantName.Should().Be(expectedName);
```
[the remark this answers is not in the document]
petrochuk: Yaml in this name is not necessary because entire library is focused on Yaml. Also control type is different from invariant template name.

src/Persistence.Tests/_TestData/ValidYaml-CI/BuiltInControl/with-template-name.pa.yaml
petrochuk: I understand the discussion about components but this PR is about PCF/non-PCF controls

PR #652, opened 2024-05-08
Yaml RedHat extension schema

petrochuk:
This is to enable simple validation for VSCode and will be expanded and refined.

### DIFF
--- a/docs/pa.yaml-schema.json
+++ b/docs/pa.yaml-schema.json
@@ -1,0 +1,53 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/microsoft/PowerApps-Language-Tooling/blob/master/docs/pa.yaml-schema.json",
+  "title": "Microsoft Power Apps canvas screen schema",
+  "description": "A schema for describing the structure of a Microsoft Power Apps Yaml.",
+  "$ref": "#/$defs/ControlDef",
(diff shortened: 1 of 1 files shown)


mizrael (approved):

joem-msft (dismissed):

docs/pa.yaml-schema.json
```diff
@@ -0,0 +1,54 @@
+{
```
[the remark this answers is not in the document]
petrochuk: This is for Yaml RedHat extension validation.

docs/pa.yaml-schema.json
```diff
@@ -0,0 +1,54 @@
...
+            "Description": { "type": "string" },
+            "Type": { "type": "string" },
+            "Screens": {
+              "type": "array",
+              "items": { "$ref": "#/$defs/ControlDef" }
+            },
+            "AccessAppScope": { "type": "string" },
+            "Settings": { "type": "object" },
+            "Style": { "type": "string" },
+            "ZIndex": { "type": "string" },
```
tehcrashxor, line 30: Should we have ZIndex here at all, since it's part of the `Properties` property, not a top level property

tehcrashxor (approved):

docs/pa.yaml-schema.json
petrochuk: Good point. Removed

PR #653, opened 2024-05-16
Setting default schema

petrochuk:


mizrael (approved):

.vscode/settings.json
```diff
@@ -9,14 +9,6 @@
   "yaml.schemaStore.enable": true,
   "yaml.validate": true,
   "yaml.schemas": {
-    "http://json-schema.org/draft-07/schema#": "*.schema.yaml",
```
joem-msft, line 12: this is the 2nd time you're deleting these other schemas. Undo the deletion of these other schemas.

.vscode/settings.json
```diff
@@ -9,14 +9,6 @@
...
-    "src/schemas/pa-yaml/v3.0/pa.schema.yaml": [
-      //"*.pa.yaml",
-      "src/schemas-tests/pa-yaml/v3.0/**/*.pa.yaml"
-    ],
-    "https://raw.githubusercontent.com/microsoft/PowerApps-Tooling/master/schemas/pa-yaml/v3.0/pa.schema.yaml": [
-      "src/schemas-tests/**/*.pa.yaml"
-    ],
-    "https://raw.githubusercontent.com/microsoft/PowerApps-Tooling/master/docs/pa.yaml-schema.json": "src/Persistence.Tests/**/*.pa.yaml"
+    "https://raw.githubusercontent.com/microsoft/PowerApps-Tooling/master/docs/pa.yaml-schema.json": "*.pa.yaml"
```
joem-msft, line 12: What is the issue that you're having?

All of the yaml files in this repo fall in the folders configured with the previous configuration. Was there a folder missing?

joem-msft (changes requested):
Please undo the removal of other schemas.

.vscode/settings.json
petrochuk: These are settings for all people who clone repo. I would like for the default settings to point to something that is working right now. We can switch to different one when other changes are finalized.

joem-msft: exactly my point here. There are two schemas in this repo, this settings.json file was configured to support both of them, along with the yaml schema files.
Bring them back so you do not break other users of this repo.

.vscode/settings.json
joem-msft: each individual should add this line to their own user-level settings.json file. This should be our recommended approach for everyone.

The configuration in here applies to just this repo and explicitly to specific file/folder paths.

It sounds like you want all other *.pa.yaml files (which do not have a schema already defined) to use the v2.2 schema correct?

may I suggest the following, with comments added for clairity:
```json
{
  "cSpell.words": [
    "Dataverse",
    "msapp",
    "PPUX",
    "RGBA"
  ],
  "[yaml]": {},
  "yaml.schemaStore.enable": true,
  "yaml.validate": true,
  "yaml.schemas": {
    // *.schema.yaml files should use JSON schema:
    "http://json-schema.org/draft-07/schema#": "*.schema.yaml",
    // This schema registration supports local development of the v3.0 schema:
    "src/schemas/pa-yaml/v3.0/pa.schema.yaml": [
      "src/schemas-tests/pa-yaml/v3.0/**/*.pa.yaml"
    ],
    // This schema registration applies the checked in master version of the schema to these folders:
    "https://raw.githubusercontent.com/microsoft/PowerApps-Tooling/master/schemas/pa-yaml/v3.0/pa.schema.yaml": [
      "src/schemas-tests/**/*.pa.yaml"
    ],
    // This schema registration is the 'default' for any *.pa.yaml files which are not covered by above registrations:
    "https://raw.githubusercontent.com/microsoft/PowerApps-Tooling/master/docs/pa.yaml-schema.json": [
      "*.pa.yaml", // Set all other pa.yaml files to default to the v2.2 schema
      // Ensure the persistence tests use v2.2 schema even if the repo default ends up changing in the future:
      "src/Persistence.Tests/**/*.pa.yaml"
    ]
  }
}
```

.vscode/settings.json
joem-msft: Please see my other comment which suggests a different config which should support what you are needing along with keeping the existing stuff working too.

swapnilbhalgat (approved):

.vscode/settings.json
petrochuk: I added some of the comments you suggested and "*.schema.yaml"

PR #654, opened 2024-05-16
Removed "Pa" prefix where it is not necessary

petrochuk:
This PR solves issue of "Pa" repeating three times.  Microsoft.PowerPlatform.**PowerApps**.Persistence.**Pa**Yaml.Serialization.**Pa**YamlSerializer

### DIFF
--- a/src/Persistence.Tests/PaYaml/Serialization/NamedObjectSequenceSerializationTests.cs
+++ b/src/Persistence.Tests/PaYaml/Serialization/NamedObjectSequenceSerializationTests.cs
@@ -10,13 +10,13 @@ namespace Persistence.Tests.PaYaml.Serialization;
 [TestClass]
 public class NamedObjectSequenceSerializationTests : SerializationTestBase
 {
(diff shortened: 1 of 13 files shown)


joem-msft (changes requested):
What is the issue you're trying to solve here? Please do not change anything in the PaYaml folders unless there is an actual feature that needs to be changed.

The PaYaml prefix in type names are by design.
- The need for a type prefix is to provide unambiguous type names in code when multiple namespaces are added via `using` statements. This is the same issue we had to address the other day.
  e.g. this assembly now has two YamlSerializer classes:
  `Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Serialization.YamlSerializer`
  and
  `Microsoft.PowerPlatform.PowerApps.Persistence.Yaml.YamlSerializer`
- The name `PaYaml` is designed to correlate with most of these files have the file extension `.pa.yaml`.

swapnilbhalgat (approved):

PR #657, opened 2024-05-23
Upgrade to .net 8

jack-work:

## Problem

.Net 7 is at end of life and we need to migrate to 8 to remain in compliance.  My ultimate intention is to upgrade document server, among this packages primary dependents, to .Net 8.
https://msazure.visualstudio.com/OneAgile/_workitems/edit/28159300/

This required me to change many files to fix code analysis errors.  Some of the errors may be controversial.  I am open to any suggestions and recommendations on what to and what not to suppress.


petrochuk:
Can we separate .net 8 upgrade from code analysis fixes?

jack-work:
Andrew did this by suppressing all of the analyzer errors.  If we want to take those up, this pr might save some time, but I don't mind if we abandon it.

PR #666, opened 2024-06-07
Copy schemas to YAML Validator and update subschema references

abaskk-msft:

## Problem

## Solution

## Changes

- The library that is going to be used for [yaml validation](https://docs.json-everything.net/schema/basics/) was not resolving the reference id's in the pa.yaml-schema.json to its subschemas (which are weblinks currently): control-type.json and control-property-schema.json. 
- I copied the schemas into the YAMLValidator project and modified the references: [here](https://github.com/microsoft/PowerApps-Tooling/blob/10f04793b06c12f6768f21118362c076dd7e0092/src/YAMLValidator/schema/pa.yaml-schema.json#L26)  and [here](https://github.com/microsoft/PowerApps-Tooling/blob/10f04793b06c12f6768f21118362c076dd7e0092/src/YAMLValidator/schema/pa.yaml-schema.json#L43) and updated the id of the subschemas to be their filenames [here](https://github.com/microsoft/PowerApps-Tooling/blob/10f04793b06c12f6768f21118362c076dd7e0092/src/YAMLValidator/schema/subschemas/control-property-schema.json#L3) and [here](https://github.com/microsoft/PowerApps-Tooling/blob/10f04793b06c12f6768f21118362c076dd7e0092/src/YAMLValidator/schema/subschemas/control-type-schema.json#L3) to allow the JSONSchema.NET library to access and store the schemas of the referenced files.
- In addition to this, I also added a ```<CopyToOutputDirectory>``` flag to copy the schemas and sub-schemas into the output directory to use relative paths.

## Validation

- How did you verify that this works: Built the YAML validator locally


petrochuk (approved):

petrochuk (approved):

src/YAMLValidator/schema/pa.yaml-schema.json
mizrael, line 1: don't copy it here, otherwise will be hard to keep it in sync. You can "link" existing files and include in the current project's build instead: https://andrewlock.net/including-linked-files-from-outside-the-project-directory-in-asp-net-core/

mizrael (changes requested):

src/YAMLValidator/schema/pa.yaml-schema.json
abaskk-msft: Can you clarify, do you mean link the existing schemas in the docs filepath from the root of the project? i.e: PowerApps-Tooling/docs/pa.yaml-schema.json

abaskk-msft: Fixed, linked the schemas and updated the original .json schemas

src/YAMLValidator/InputProcessor.cs
```diff
@@ -54,7 +54,7 @@ public static RootCommand GetRootCommand()
         var schemaOption = new Option<string>(
             name: "--schema",
             description: "The path to the schema json file",
-            getDefaultValue: () => @"..\schemas\pa-yaml\v3.0\pa.yaml-schema.json"
+            getDefaultValue: () => @".\schema\pa.yaml-schema.json"
```
mizrael, line 57: nit: I'd keep the version number

docs/control-property-schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://raw.githubusercontent.com/microsoft/PowerApps-Tooling/master/docs/control-type-schema.json",
```
mizrael, line 3: was it necessary to remove the full url?

abaskk-msft: Yes. Id's help the  [json-everything library](https://github.com/gregsdennis/json-everything) resolve references to local subschema files that are referenced in the [pa.yaml-schema.json](https://github.com/microsoft/PowerApps-Tooling/blob/f662d459b9d542eb26be6bd4043da42ea02430c8/docs/pa.yaml-schema.json#L26) file

abaskk-msft: Alternatively, if a user wants to reference source schema in the repo, we could add a [json-schema comment](https://json-schema.org/understanding-json-schema/reference/comments)

src/YAMLValidator/InputProcessor.cs
abaskk-msft: just to clarify, I put that as a dummy filepath since this was just the scaffolding. The actual .json schema files are from the [docs folder from the root of the repo](https://github.com/microsoft/PowerApps-Tooling/tree/master/docs)

mizrael: I know, my was just a suggestion.

PR #669, opened 2024-06-13
Skip ZIndex from being serialized.

shshetty-microsoft:

## Problem

Currently when we vie code for gallery, the ZIndex is being serialized at the parent level. Changes here are the skip the ZIndex

## Solution

Add a check to skip Zindex

## Changes

- Log what was changed
- Log what was changed
- Log what was changed

## Validation

### DIFF
--- a/src/Persistence/Yaml/ControlFormatter.cs
+++ b/src/Persistence/Yaml/ControlFormatter.cs
@@ -20,7 +20,7 @@ public static class ControlFormatter
         _ = control ?? throw new ArgumentNullException(nameof(control));
 
         var childrenToRemove = (control.Children ?? Enumerable.Empty<Control>()).Where(c => c.Template.AddPropertiesToParent).ToList();
-        var propertiesToMerge = childrenToRemove.SelectMany(c => c.Properties).ToList();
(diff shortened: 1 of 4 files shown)


src/Persistence/Yaml/ControlFormatter.cs
```diff
@@ -20,7 +20,7 @@ public static class ControlFormatter
         _ = control ?? throw new ArgumentNullException(nameof(control));
 
         var childrenToRemove = (control.Children ?? Enumerable.Empty<Control>()).Where(c => c.Template.AddPropertiesToParent).ToList();
-        var propertiesToMerge = childrenToRemove.SelectMany(c => c.Properties).ToList();
+        var propertiesToMerge = childrenToRemove.SelectMany(c => c.Properties.Where(p => p.Key != PropertyNames.ZIndex)).ToList();
```
tehcrashxor, line 23: This is already done below on line 40, and the ZIndex still needs to be present for the sorting code on lines 32-35 prior to that removal.

tehcrashxor: or wait, this is the ZIndex of the children with AddPropertiesToParent flag, not this current control's ZIndex.   
Nevermind.

tehcrashxor (approved):

petrochuk (approved):

PR #674, opened 2024-06-21
Fix broken build, and inner dev-loop configurations

joem-msft:
This PR re-fixes issues in the inner dev loop which I'd previously fixed but have been broken again. All of these changes have been fixed at one point, hopefully it'll stick this time.

- Turned off TreatWarningsAsErrors in VS IDE to speed up inner-dev-loop.
   building locally in VS IDE causes warnings to be treated as Errors. This negatively impacts the inner dev loop experience and causes slower iterative development.
- fix settings.json so PaYamlV3 files use the correct schema file
  *.pa.yaml files under the PaYamlV3 folders in the repo no longer work with the schemas as they did before.
- Turned some code analysis rules from Error to Warning to improve inner dev-loop

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -120,7 +120,7 @@ dotnet_code_quality_unused_parameters = all:suggestion
 [*.cs]
 csharp_indent_labels = one_less_than_current
 csharp_using_directive_placement = outside_namespace:error
-csharp_prefer_simple_using_statement = true:error
+csharp_prefer_simple_using_statement = true:suggestion
 csharp_prefer_braces = true:silent
 csharp_style_namespace_declarations = file_scoped:error
 csharp_style_prefer_method_group_conversion = true:silent
@@ -188,7 +188,7 @@ dotnet_analyzer_diagnostic.category-Naming.severity           = warning
 dotnet_analyzer_diagnostic.category-Performance.severity      = warning
 dotnet_analyzer_diagnostic.category-Reliability.severity      = warning
 dotnet_analyzer_diagnostic.category-Security.severity         = error
-dotnet_analyzer_diagnostic.category-Style.severity            = error
+dotnet_analyzer_diagnostic.category-Style.severity            = warning
 dotnet_analyzer_diagnostic.category-Usage.severity            = warning
 
 # *********************************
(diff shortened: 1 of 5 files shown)


mizrael (approved):

.editorconfig
```diff
@@ -120,7 +120,7 @@ dotnet_code_quality_unused_parameters = all:suggestion
 [*.cs]
 csharp_indent_labels = one_less_than_current
 csharp_using_directive_placement = outside_namespace:error
-csharp_prefer_simple_using_statement = true:error
```
petrochuk, line 123: Can we keep this as an error?

.editorconfig
```diff
@@ -188,7 +188,7 @@ dotnet_analyzer_diagnostic.category-Naming.severity           = warning
 dotnet_analyzer_diagnostic.category-Performance.severity      = warning
 dotnet_analyzer_diagnostic.category-Reliability.severity      = warning
 dotnet_analyzer_diagnostic.category-Security.severity         = error
-dotnet_analyzer_diagnostic.category-Style.severity            = error
+dotnet_analyzer_diagnostic.category-Style.severity            = warning
```
petrochuk, line 191: Can we keep this as an error?

.editorconfig
```diff
@@ -200,9 +200,10 @@ dotnet_analyzer_diagnostic.category-Usage.severity            = warning
 
 dotnet_diagnostic.IDE0003.severity = error      # this and Me preferences
 dotnet_diagnostic.IDE0007.severity = error      # Use 'var' instead of explicit type
-dotnet_diagnostic.IDE0005.severity = error      # Remove unnecessary using directives
+dotnet_diagnostic.IDE0005.severity = warning    # Remove unnecessary using directives
```
petrochuk, line 203: Let's keep this as an error too

petrochuk (changes requested):

.editorconfig
tehcrashxor: This should be fine with Joe's change.  We still have warnings-as-errors for everything but the local VS IDE build, so setting it to warning here should speed up the inner loop while still being errors before check-in.

.editorconfig
tehcrashxor: This should be fine with Joe's change.  We still have warnings-as-errors for everything but the local VS IDE build, so setting it to warning here should speed up the inner loop while still being errors before check-in.

.editorconfig
tehcrashxor: While the simplified using is nicer in 95%+ places, anywhere that we want to explicitly make sure the `.Dispose()` is called prior to continuing still needs the old syntax.
E.g., doing file operations, but needing the using's `.Dispose()` to close and flush the stream before opening a second time in the same method.

tehcrashxor (approved):

swapnilbhalgat (approved):

PR #675, opened 2024-06-21
Extend MsappArchive APIs

joem-msft:
This PR moves some new APIs used in the DocSvr code into the MsappArchive and IMsappArchive APIs in this repo.
- `DoesEntryExist` - allows callers to determine whether an entry exists w/o needing to inspect internal canonical dictionary.
- `TryGenerateUniqueEntryPath` - exposes logic for generating a unique entry, in case similar entries already exist. This function is similar to the private method `GetSafeEntryPath`, but this method only has a single responsibility, in order to make it more usable for different scenarios.
- `MsappArchive.TryMakeSafeForEntryPathSegment` is a new static method that has a single responsibility, allowing it's usage to be composed with other functions  depending on the scenario. It is similar to encoding a path segment of a URI. Namely, the char set it allows is anything that could be a valid file name or directory name.
- `GetSafeEntryPath` is modified to use the new `TryGenerateUniqueEntryPath` and `TryMakeSafeForEntryPathSegment` methods.
- The `GetEntry`, `GetRequiredEntry` methods were updated to utilize the `TryGetEntry` method.

Validation:
- Added additional unit tests for the three new methods
- Updated existing test output due to '.' chars now being a valid file/directory name.


mizrael (commented):
do we really need to add these methods to the interface? can't we have them exposed as extension methods instead?

src/Persistence.Tests/MsApp/MsappArchiveSaveTests.cs
```diff
@@ -11,9 +11,9 @@ public class MsappArchiveSaveTests : TestBase
 {
     [TestMethod]
     [DataRow(@"  Hello   ", $"src/Hello.pa.yaml", @"_TestData/ValidYaml-CI/Screen-Hello1.pa.yaml")]
-    [DataRow(@"..\..\Hello", $"src/Hello.pa.yaml", @"_TestData/ValidYaml-CI/Screen-Hello2.pa.yaml")]
-    [DataRow(@"c:\win\..\..\Hello", $"src/cWinHello.pa.yaml", @"_TestData/ValidYaml-CI/Screen-Hello3.pa.yaml")]
-    [DataRow(@"//..?HelloScreen", $"src/HelloScreen.pa.yaml", @"_TestData/ValidYaml-CI/Screen-Hello4.pa.yaml")]
+    [DataRow(@"..\..\Hello", $"src/....Hello.pa.yaml", @"_TestData/ValidYaml-CI/Screen-Hello2.pa.yaml")]
```
petrochuk, line 14: Can we keep existing test cases and add new ones instead?

src/Persistence.Tests/MsApp/MsappArchiveTests.cs
```diff
@@ -4,12 +4,25 @@
 using System.IO.Compression;
 using Microsoft.PowerPlatform.PowerApps.Persistence;
 using Microsoft.PowerPlatform.PowerApps.Persistence.MsApp;
+using Microsoft.PowerPlatform.PowerApps.Persistence.Yaml;
+using Moq;
```
petrochuk, line 8: It is not recommended to use Moq because of serious privacy concerns: https://github.com/devlooped/moq/issues/1372

src/Persistence/MsApp/MsappArchive.cs
```diff
@@ -452,24 +455,49 @@ public void Save(Control control, string? directory = null)
...
+        {
+            throw new ArgumentException("The directory can be null, but cannot be empty or whitespace only.", nameof(directory));
+        }
+        _ = !string.IsNullOrEmpty(fileNameNoExtension) ? fileNameNoExtension : throw new ArgumentNullException(nameof(fileNameNoExtension));
+
+        var entryPathPrefix = directory == null ? fileNameNoExtension : Path.Combine(directory, fileNameNoExtension);
+
+        // First see if we can use the name as is
+        entryPath = $"{entryPathPrefix}{extension}";
+        if (!DoesEntryExist(entryPath))
```
petrochuk, line 487: This checks raw path against canonical entries which can be an issue if path is "/entryA" for example

petrochuk (changes requested):

tehcrashxor (approved):

tehcrashxor (approved):

swapnilbhalgat (approved):

PR #677, opened 2024-06-24
Yaml Validator Iteration One into Main Branch

abaskk-msft:
## Feature
- This feature adds a Yaml Validator under the Persistence namespace. The inputs accepted are either a single yaml file or a directory of yaml files. Here is a rough UML of the structure of the project.
- Additionally users are able to see a stack trace of syntax errors. Specifically the file where an error was found and where the exact error with yaml pointer and schema pointer paths.

image

## Validation
- Tested the VSCode extension still works despite the changes in schema format
- Unit tested the Validator class


src/YAMLValidator.Tests/ValidatorTest.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Json.Schema;
+using Microsoft.PowerPlatform.PowerApps.Persistence;
```
petrochuk, line 5: Please add "YamlValidatorTests" so it becomes: Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidatorTests;

src/YAMLValidator/InputProcessor.cs
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.CommandLine;
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence;
```
petrochuk, line 6: Please add "YAMLValidator" so it becomes Microsoft.PowerPlatform.PowerApps.Persistence.YAMLValidator;

petrochuk (approved):

petrochuk (approved):

src/YAMLValidator/Constants.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+public class Constants
+{
+    public const string FileTypeName = "file";
+    public const string FolderTypeName = "folder";
+    public const string YamlFileExtension = ".yaml";
+    public const string YmlFileExtension = ".yml";
+    public const string JsonFileExtension = ".json";
+
+    public const string verbose = "verbose";
```
tehcrashxor, line 13: Style nitpick: `public const` and `public static readonly` fields should start upper case:
```suggestion
    public const string Verbose = "verbose";
```

docs/pa.yaml-schema.json
```diff
@@ -23,7 +23,7 @@
           "type": "object",
           "properties": {
             "Control": {
-              "$ref": "control-type-schema.json#"
```
petrochuk, line 26: This change breaks VSCode schema validation in my tests

petrochuk (changes requested):

src/YAMLValidator/Utility.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// using YamlDotNet.Core;
```
tehcrashxor, line 4: In general, we should remove commented out code
```suggestion
```

Some exceptions for work in progress, or awaiting some other change, may be fine, but should typically have an explanatory comment on why they're commented out.

src/YAMLValidator/YAMLValidator.csproj
```diff
@@ -0,0 +1,35 @@
...
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+      <NoWarn>1591, CA1822</NoWarn>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+      <NoWarn>1591, CA1822</NoWarn>
+    </PropertyGroup>
```
tehcrashxor, line 16: Any `<NoWarn>` in the project file should typically have a comment explaining what warnings are being suppressed.  
They're fairly simple to look up, but it's simpler for readers (either in PR or looking over the file in the future) to see it explicitly named.

petrochuk (approved):

src/YAMLValidator/InputProcessor.cs
```diff
@@ -0,0 +1,110 @@
...
+        var validateCommand = new Command("validate", "Validate the input yaml file")
+        {
+            pathOption,
+            schemaOption
+        };
+
+        validateCommand.SetHandler((pathOptionVal, schemaOptionVal) =>
+        {
+            // validation has completed, we either have a file or folder
+            var pathType = File.Exists(pathOptionVal) ? Constants.FileTypeName :
```
cwduncan, line 99: This is a more explicit way to validate if a path leads to a folder:
`File.GetAttributes(filePath).HasFlag(FileAttributes.Directory);`

petrochuk (approved):

src/YAMLValidator/InputProcessor.cs
abaskk-msft: added and fixed this

src/YAMLValidator/YAMLValidator.csproj
abaskk-msft: I added an explanation as to why CA1822 was added, the other error was previously present from the Directory.Build.Props

src/YAMLValidator/Constants.cs
abaskk-msft: fixed

src/YAMLValidator/Utility.cs
abaskk-msft: just a leftover from something I tried so I ended up deleting this . I did add a comment to this based on what you suggested https://github.com/microsoft/PowerApps-Tooling/blob/afaf6d739d9152483f6040a1a9ee84d552ac0a9b/src/YamlValidator/Validator.cs#L38 makes, sense this was

tehcrashxor (approved):

PR #679, opened 2024-06-25
PaYaml V3 updates

joem-msft:
- add `ScreenInstance.Variant`
- `PaYamlSerializationContext` now holds the `PaYamlSerializerOptions` options instance, and moved YamlDotNet builder configs to this class. This helps to reduce the number of parameters needed for some custom converter operations, along with keeping the options class more like a POCO.
- Used primary ctor in some locations
- PaYaml OM properties for collection properties (`Properties`, `Children`, `Groups`) are now nullable, default to null, to conserve memory all around.
- Reverted type name changes so the term `PaYaml` is preferred over just `Yaml`. As was already discussed as part of original changes, this term is used to differentiate the types from more general purpose Yaml serialization types.


tehcrashxor (approved):

src/Persistence.Tests/PaYaml/Models/PaControlInstanceContainerTests.cs
```diff
@@ -14,12 +14,11 @@ public void DescendantControlInstancesSingleLevel()
...
         var screen = new NamedObject<ScreenInstance>("Screen1", new()
         {
-            Children =
-            {
+            Children = [
                 new("Ctrl0", new("GroupContainer")),
                 new("Ctrl1", new("GroupContainer")),
                 new("Ctrl2", new("GroupContainer")),
-            },
+                ],
```
swapnilbhalgat, line 21: nit: is this correct indentation?

swapnilbhalgat (approved):

PR #682, opened 2024-06-26
Clean yaml validator false positive output

abaskk-msft:
## Problem

- In the current validation output there is an edge case with returning output on an Invalid Schema.
- The schema structure requires that at the root of the yaml we either have a single control or an array of controls (using a OneOf json-schema statement)
image
- In a valid schema, a single evaluation path being true (object or array of objects) will make the entire schema true
- In an invalid schema, both evaluation paths will be false. Suppose we wanted to trace an error; we would have to go down both evaluation paths in the evaluation tree (one of which would always evaluate to false even if the schema was valid). 
- This additional useless information is output on the console to the user and is confusing/uninformative. This pr removes the evaluation path which contains useless information.

## Solution

- Among the schema evaluation path which require we have a single control object or array of control object; we find which evaluation path has the longest suffix and filter out the rest of the evaluation paths with a smaller length suffix. 

## Validation

- Only local test with examples I made. Perhaps we can add unit tests on the actual errors of the schema, as opposed to simply checking if the entire schema evaluated true or false

### DIFF
--- a/src/YamlValidator/ValidatorResults.cs
+++ b/src/YamlValidator/ValidatorResults.cs
@@ -12,11 +12,11 @@ public class ValidatorResults
     public ValidatorResults(bool schemaValid, IReadOnlyList<ValidatorError> traversalResults)
     {
         SchemaValid = schemaValid;
-        TraversalResults = filterErrors(traversalResults);
+        TraversalResults = FilterErrors(traversalResults);
     }
 
     //  This will filter out the false positives that are not relevant to the error output, when the validation is false
-    private ReadOnlyCollection<ValidatorError> filterErrors(IReadOnlyList<ValidatorError> traversalResults)
+    private ReadOnlyCollection<ValidatorError> FilterErrors(IReadOnlyList<ValidatorError> traversalResults)
     {
         var maxSchemaArraySuffixSize = 0;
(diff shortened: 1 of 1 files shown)


petrochuk (approved):

src/YamlValidator/ValidatorResults.cs
```diff
@@ -10,6 +12,63 @@ public class ValidatorResults
     public ValidatorResults(bool schemaValid, IReadOnlyList<ValidatorError> traversalResults)
     {
         SchemaValid = schemaValid;
-        TraversalResults = traversalResults;
+        TraversalResults = filterErrors(traversalResults);
+    }
+
+    //  This will filter out the false positives that are not relevant to the error output, when the validation is false
+    private ReadOnlyCollection<ValidatorError> filterErrors(IReadOnlyList<ValidatorError> traversalResults)
```
petrochuk, line 19: nit: Methods usually start with capital letter

petrochuk (approved):

src/YamlValidator/ValidatorResults.cs
abaskk-msft: Did this out of habit ☹️ . Fixed now

cwduncan (approved):

PR #684, opened 2024-06-28
Return Formulas.Tools to netstandard2.0

tehcrashxor:
In commit [1c9b22c](https://github.com/microsoft/PowerApps-Tooling/commit/1c9b22cef15b0da9d8afbdb57aebaac3252807ed) back in Dec 2023, all projects in the repo were set to use the same TargetFramework, all sourced from a single shared file.

This accidentally effected the Formulas.Tooling lib, which needs to remain on netstandard2.0 for use in both .NET and .NET Framework versions of PAC CLI.

This PR switches that library back to netstandard2.0, and reverts a few changes since that point for new syntax net8.0 which is not available to netstandard2.0.

### DIFF
--- a/src/PAModel/Microsoft.PowerPlatform.Formulas.Tools.csproj
+++ b/src/PAModel/Microsoft.PowerPlatform.Formulas.Tools.csproj
@@ -1,6 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>12</LangVersion>
     <AssemblyName>Microsoft.PowerPlatform.Formulas.Tools</AssemblyName>
     <RootNamespace>Microsoft.PowerPlatform.Formulas.Tools</RootNamespace>
   </PropertyGroup>
@@ -48,9 +50,5 @@
     <PackageReference Include="System.Text.Json" Version="[6.0,)" />
     <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0" />
     <PackageReference Include="YamlDotNet" Version="$(YamlDotNetVersion)" />
-  </ItemGroup>
-    
-  <ItemGroup>
-    <ProjectReference Include="..\Persistence\Microsoft.PowerPlatform.PowerApps.Persistence.csproj" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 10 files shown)


petrochuk (approved):

src/PAModel/Microsoft.PowerPlatform.Formulas.Tools.csproj
```diff
@@ -49,8 +51,4 @@
     <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0" />
     <PackageReference Include="YamlDotNet" Version="$(YamlDotNetVersion)" />
   </ItemGroup>
-    
-  <ItemGroup>
-    <ProjectReference Include="..\Persistence\Microsoft.PowerPlatform.PowerApps.Persistence.csproj" />
```
joem-msft, line 54: oh nice! glad the dependency isn't actually necessary.

src/PAModel/Serializers/MsAppSerializer.cs
```diff
@@ -57,7 +57,10 @@ private static string AsString(ZipArchiveEntry entry)
 
     public static CanvasDocument Load(Stream streamToMsapp, ErrorContainer errors)
     {
-        ArgumentNullException.ThrowIfNull(streamToMsapp);
+        if (streamToMsapp == null)
+        {
+            throw new ArgumentNullException(nameof(streamToMsapp));
```
joem-msft, line 62: nit: `_ = streamToMsapp ?? throw new ArgumentNullException(nameof(streamToMsapp));`

src/PAModel/Microsoft.PowerPlatform.Formulas.Tools.csproj
```diff
@@ -1,6 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
joem-msft, line 4: consider adding `net8.0` as a target (change to plural: `TargetFrameworks` even when there's only one).
this is good practice for allowing support of older frameworks, but to jump forward to latest .net when consumers support it.

joem-msft: it means that doc svr will be able to utilize the net8.0 version of the dll

joem-msft (approved):

src/PAModel/Microsoft.PowerPlatform.Formulas.Tools.csproj
tehcrashxor: Good idea.  Done.

src/PAModel/Serializers/MsAppSerializer.cs
tehcrashxor: done

petrochuk (approved):

PR #687, opened 2024-06-28
Make YamlValidator filepaths platform agnostic

abaskk-msft:
## Problem

- The repo is meant to be platform agnostic, the YamlValidator isn't. This PR attempts to make code involving filepaths and their test platform agnostic.

## Notes
- Tests became more verbose since the filepath type and directory path is determined at runtime and had to be moved out of DataRows.

## Validation

- re-ran unit tests 
- Tested Locally


src/YamlValidator/Constants.cs
```diff
@@ -11,4 +11,8 @@ public static class Constants
     public const string JsonFileExtension = ".json";
 
     public const string Verbose = "verbose";
+
+    // runtime constants
+    // default schema path
+    public static readonly string[] DefaultSchemaPath = [".", "schema", "pa.yaml-schema.json"];
```
tehcrashxor, line 17: You can also just do the `Path.Combine(...)` here directly:
```suggestion
    public static readonly string DefaultSchemaPath = Path.Combine(".", "schema", "pa.yaml-schema.json");
```

src/YamlValidator.Tests/ValidatorTest.cs
```diff
@@ -9,38 +9,49 @@ namespace Persistence.Tests.YamlValidator;
 [TestClass]
 public class ValidatorTest
 {
-    private const string _validPath = @".\_TestData\ValidYaml";
-    private const string _invalidPath = @".\_TestData\InvalidYaml";
-    private const string _schemaPath = @"..\YamlValidator\schema\pa.yaml-schema.json";
+
+    private static readonly string _validPath = Path.Combine(
+        [".", "_TestData", "ValidYaml"]) + Path.DirectorySeparatorChar;
```
tehcrashxor, line 14: This overload of `Path.Combine(...)` takes a `params` argument, so no need to create an array yourself here - you can remove the `[` and `]` and the compiler handles the array creation:
```suggestion
    private static readonly string _validPath = Path.Combine(".", "_TestData", "ValidYaml") + Path.DirectorySeparatorChar;
```

tehcrashxor (approved):

petrochuk (approved):

PR #692, opened 2024-07-02
Yaml Validator Bug: same file names in different folder cause error

abaskk-msft:
## Problem
- File names are stored in a dictionary so that they can be output in the console
- Two of the same file names in different folders cause an error when recursively traversing the folders due to a repeated key (was found when traversing the Persistence.Tests Valid-CI test data folder)

## Changes
The way file names are stored in a dictionary and output. Specifically, use the relative path of a file as the name of the file instead of just the filename.

## Validation
- Re-ran unit tests, manual validation

### DIFF
--- a/src/YamlValidator/YamlLoader.cs
+++ b/src/YamlValidator/YamlLoader.cs
@@ -25,7 +25,7 @@ public class YamlLoader
                 var yamlFiles = Directory.EnumerateFiles(filePath, "*" + Constants.YamlFileExtension, SearchOption.AllDirectories);
                 foreach (var filename in yamlFiles)
                 {
-                    var fileName = Path.GetFileName(filename);
+                    var fileName = Path.GetRelativePath(filePath, filename);
                     var yamlText = Utility.ReadFileData(filename);
                     deserializedYaml.Add(fileName, yamlText);
                 }
--- a/src/YamlValidator/YamlLoader.cs
+++ b/src/YamlValidator/YamlLoader.cs
@@ -25,7 +25,7 @@ public class YamlLoader
                 var yamlFiles = Directory.EnumerateFiles(filePath, "*" + Constants.YamlFileExtension, SearchOption.AllDirectories);
                 foreach (var filename in yamlFiles)
                 {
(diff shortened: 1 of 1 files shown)


petrochuk (approved):

src/YamlValidator/YamlLoader.cs
```diff
@@ -25,7 +25,7 @@ public class YamlLoader
                 var yamlFiles = Directory.EnumerateFiles(filePath, "*" + Constants.YamlFileExtension, SearchOption.AllDirectories);
                 foreach (var filename in yamlFiles)
                 {
-                    var fileName = Path.GetFileName(filename);
+                    var fileName = Path.GetRelativePath(filePath, filename);
```
mizrael, line 28: I'm not 100% sure, but I think https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfullpath?view=net-8.0#system-io-path-getfullpath(system-string-system-string) would be a better choice

mizrael (changes requested):

src/YamlValidator/YamlLoader.cs
abaskk-msft: My thinking in using relative paths was that the users wouldn't have to deal with verbose file paths (they input the folder path so the relative paths should be straightforward). In the CI, this would mean we wouldn't get an uninformative longer path from the container's absolute file path.

abaskk-msft: I think using the full path is much safer though so I guess we can go with that

abaskk-msft: Can you elaborate? As in use a full directory path? 
> Using a predefined basePath rather than one based on the current drive directory guards against unwanted file paths caused by unexpected changes in the current drive and directory. (from the docs)
We would naturally expect the user to not change the location of the path they input when running the cli tool

mizrael: correct. I'd suggest using the complete absolute path to the file as dictionary key.

cwduncan (approved):

PR #694, opened 2024-07-03
Integrate Yaml Validator into the CI to validate Persistence Test cases

abaskk-msft:
## Feature

- Further validation of the test case Yaml Files in Persistence.Tests using the YamlValidator to run on our CI pipeline
- To do: add CI feature flag to CLI tool to create Github actions warnings on files which don't pass validation instead of simply outputting results to console
- To do: use environment variables to set library versions.

## Changes
- Removed directory validation in the CLI that would error if there were no pa.yaml files in the directory, since we now recursively traverse directories.
- Uses same MSTest library as Persistence (which includes the test runner) so that YamlValidator unit tests actually run during CI.
- Github CI workflow modification
- Addition of powershell and bash scripts

## Validation
- Unit test, and manual run
- Test CI runs using github (as evidenced by the number of commits)😄


mizrael:
I think the plan is to move the validation logic inside the Persistence library first. At that point we can get rid of the console app, since the next step would be updating PAC CLI.
Regarding our tests suite, once the Persistence library has been updated, we can write tests that make use of the new validation code and run on specific folders. This way we can run them directly via `dotnet test`, perhaps filtering them via a traits/categories attributes.

.github/workflows/CI.yml
```diff
@@ -40,12 +40,23 @@ jobs:
       run: ./build ci
       if: ${{ runner.os != 'Windows' }}
 
-    - name: Test - Persistence
-      run: dotnet bin/Debug/Persistence.Tests/Persistence.Tests.dll
+    - name: Test - YamlValidator
```
joem-msft, line 43: could we move the validator tests to come after the core output libraries get tested?
If the base assemblies fail, then there may be no reason to also run tests on clients that depend on those libraries.

.github/workflows/CI.yml
```diff
@@ -40,12 +40,23 @@ jobs:
...
 
+    - name: Test - Persistence.Tests Yaml Files (ubuntu & macos)
+      run: ./scripts/PersistenceValidate.sh
+      if: ${{ runner.os != 'Windows' }}
+
+    - name: Test - Persistence.Tests Yaml Files (windows)
+      run: ./scripts/PersistenceValidate.ps1
+      if: ${{ runner.os == 'Windows' }}
+
+    - name: Test - Persistence
```
joem-msft, line 57: move these dll tests before the Yaml file tests

joem-msft (changes requested):

.github/workflows/CI.yml
```diff
@@ -46,6 +46,17 @@ jobs:
     - name: Test - PAModel
       run: dotnet bin/Debug/PAModelTests/PAModelTests.dll
 
+    - name: Test - YamlValidator
+      run: dotnet bin/Debug/YamlValidator.Tests/YamlValidator.Tests.dll
+
+    - name: Test - Persistence.Tests Yaml Files (ubuntu & macos)
```
joem-msft, line 52: note: you can simplify cross-platform scripting if you use PowerShell Core scripts. it makes it much simpler to implement and maintain when there's only one scripting language.

consider this for your next PR.

joem-msft (approved):

petrochuk (approved):

PR #700, opened 2024-07-15
Move Yaml Validator into Persistence Part 2

abaskk-msft:
## PR Update
- Move Yaml validator out of Persistence again with the existing refactoring updates. As Persistence uses .net8 , while pac supports several other .net verisons.
- The intention is to still package YamlValidator into a separate NuGet package from Persistence (due to compatbility issues with PAC CLI). The build process of signing this binary and adding it to the build pipeline will be done in a separate pr.
## Problem
- At it's current state, the YamlValidator can't be properly consumed by pac CLI to perform validation
- This re-write fixes that with a factory class to build validators (which can be extended if need be in the future)

## Notes
- The factory for the Validator is barebones, however some extensions that can be made using the factory include:
	- passing in options to produce errors in different languages: https://docs.json-everything.net/schema/basics/#schema-error-localization
	- Passing in parameters to output raw json schema evaluation results to feed into other programs or for internal use
	- passing in parameters to clarify schema spec (ex: 2020-12) used by json schema
- Error handling:
	- The Validator will gracefully handle any errors that involve edge cases when validating yaml strings (empty, not yaml etc) as well as any issues with the schema in the nuget package
- File I/O error will always be handled by pac cli

image

## Validation
- Unit tests
- Added unit test for factory class as well

## Schema Loading
- Instead of using file I/O and building to a local directory, schemas are built into the assembly itself , subschemas and schemas are then loaded into memory, like bolt.module.connector does


src/Persistence/Microsoft.PowerPlatform.PowerApps.Persistence.csproj
```diff
@@ -32,7 +32,8 @@
     <PublicSign>true</PublicSign>
     <!-- Workaround for version range https://github.com/NuGet/Home/issues/11842 -->
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
-    <NoWarn>$(NoWarn);NU1601</NoWarn>
+    <!-- YamlValidator Supressions: CA1822, IDE0052 (unused json serializer) in YamlValidator -->
+    <NoWarn>$(NoWarn);NU1601;CA1822;IDE0052</NoWarn>
```
joem-msft, line 36: undo these suppressions.
Instead, either address the diagnostics, or suppress them in one of the many ways. Often, I prefer using the attribute method for things. But you can also update the .editorconfig

src/Persistence/YamlValidator/SchemaLoader.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+
+public class SchemaLoader
+{
+    private const string _schemaFolderPath = "schema";
+    private const string _subschemaFolderPath = "subschemas";
+
+    public JsonSchema Load()
+    {
+        var assembly = Assembly.GetExecutingAssembly();
```
joem-msft, line 13: due to performance issues, do not use `Assembly.GetExecutingAssembly()`. When getting types etc, use `typeof(SomeType)` when the type is known statically, or `value.GetType()` when only known at runtime.

joem-msft: you can use `typeof(SomeType).Assembly`

src/Persistence/YamlValidator/Validator.cs
```diff
@@ -4,40 +4,45 @@
 using Json.Schema;
 using Yaml2JsonNode;
 using System.Text.Json;
+using YamlDotNet.Core;
```
joem-msft, line 7: in order to isolate dependents from also having to take an active dependency on YamlDotNet and Yaml2JsonNode, etc, be sure that you do not expose any of their types in the public interfaces for this assembly.

For example, we don't want the DocumentServer or Pac CLI to take a dependency on these 3rd party assemblies, as we may change our own dependencies on them over time. We've done this by making sure that we never actually expose any YamlDotNet types in this assembly's public API surface. We create an interface to wrap certain things, and factories etc. Other options is to create our own 'Options' pattern records for capturing the options we want to expose, but use them to configure the YamlDotNet options at runtime.

The same pattern should be done with these new Yaml2JsonNode types.

Where is the `EvaluationOptions` type defined? and the `JsonSchema` type?

One technique to do this would be to create `IValidatorFactory` and `IValidator` interfaces. Then make the `Validator` and `ValidatorFactory` classes have `internal` visibility.

joem-msft (changes requested):
great progress. Please address the comments that I've raised.

src/Persistence/YamlValidator/Validator.cs
abaskk-msft: > Where is the EvaluationOptions type defined? and the JsonSchema type?
Bot of these types are defined [here](https://github.com/json-everything/json-everything) -> this library is the one used for validation

src/Persistence/YamlValidator/ValidationProcessor.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+
+public class ValidationProcessor
+{
+    private readonly YamlLoader _fileLoader;
+    private readonly SchemaLoader _schemaLoader;
+    private readonly Validator _validator;
+
+    public ValidationProcessor(YamlLoader fileLoader, SchemaLoader schemaLoader, Validator validator)
```
joem-msft: should this use the new IValidator interface?

src/Persistence/YamlValidator/ValidationRequest.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+
+public readonly record struct ValidationRequest(string FilePath, string FilePathType);
```
joem-msft: what does the `FilePathType` mean? a name like this often means an enum type would be good. but it's not clear from usage.

src/Persistence/YamlValidator/Validator.cs
```diff
@@ -9,9 +9,11 @@
 
 namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
 
-public class Validator
+internal class Validator : IValidator
 {
     private readonly EvaluationOptions _verbosityOptions;
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("CodeQuality", "IDE0052:Remove unread private member",
```
joem-msft, line 15: don't suppress. remove it until it's actually needed.

src/Persistence/YamlValidator/ValidatorFactory.cs
```diff
@@ -6,9 +6,11 @@
 
 namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
 
-public class ValidatorFactory
+internal class ValidatorFactory : IValidatorFactory
 {
-    public Validator GetValidator()
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1822:Mark members as static",
```
joem-msft, line 11: I personally feel like this should be a suggestion, and not enforced. but there are performance reasons for it.
But, I believe a better Justification would be "required by IValidatorFactory interface"

src/Persistence/YamlValidator/IValidatorFactory.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+
+public interface IValidatorFactory
+{
+    IValidator GetValidator();
```
joem-msft, line 8: 'Get' conveys that the thing being retrieved already exists. member data or db data lookup.
'Create' conveys a new instance will always be returned.
There are some exceptions, but from the consumer of the interface, which is a Factory pattern, the expectation is that a new instance should get constructed. So rename to 'CreateValidator'.

src/Persistence/YamlValidator/Validator.cs
abaskk-msft: I went with your suggestion of making the `Validator.cs` and `ValidatorFactory.cs` internal as well as making public interfaces for them. In addition to this I applied your suggestion of not exposing the 3rd party assembly types [here](https://github.com/microsoft/PowerApps-Tooling/blob/82f1abbba7bff3e01330015e7af7bf49b55dae23/src/Persistence/YamlValidator/ValidatorError.cs)  in the encapuslated type I made, ValidatorError.cs. Now, instead of accepting a type from the Json.Schema library (EvaluationResults) type it now uses strings, and a dictionary when being instantiated.

src/Persistence/YamlValidator/ValidationProcessor.cs
abaskk-msft: This file was intended to be deleted. These are results of the merge conflict (which should now be fixed). I wanted to re-request review after I fixed the merge conflict so the PR wouldn't be confusing 😞

src/Persistence/YamlValidator/ValidationRequest.cs
abaskk-msft: This is the result of the merge conflict. This is unused and should be removed now

.version/PipelineAssemblyInfo.cs
```diff
@@ -3,6 +3,8 @@
 
 // This is pipeline generated file. Do not modify. This will be replaced with the actual versions in the actual Pipeline.
 using System.Reflection;
+using System.Runtime.CompilerServices;
 [assembly: AssemblyVersion("0.0.0.0")]
 [assembly: AssemblyFileVersion("0.0.0.0")]
-[assembly: AssemblyInformationalVersion("0.0.0.0-dev-00000000")]
\ No newline at end of file
+[assembly: AssemblyInformationalVersion("0.0.0.0-dev-00000000")]
+[assembly: InternalsVisibleTo("Persistence.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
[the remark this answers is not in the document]
abaskk-msft: Makes sense. To clarify, do you mean adding an AssemblyInfo.cs into the Properties folder in Persistence and then adding the internals visible to there?

src/Persistence/YamlValidator/Validator.cs
abaskk-msft: Another question: since the factory's implementation is internal too wouldn't I need a public provider to instantiate the factory ? More specifically: say I want to use `Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;` in PAC cli, instantiating the factory would cause an error. So instead I use a public provider which instantiates and returns a ValidatorFactory interface

petrochuk (approved):

tehcrashxor (approved):

PR #704, opened 2024-07-23
Changes to address gallery style issue.

shshetty-microsoft:

## Problem

In the case of the gallery control, there is a nested template that needs to be available before the children control for gallery are created. The nested template is used to provide information about the data control that the gallery needs to use. Changes to add nested children first. Validate change in test case.

## Solution

Add the nested templates before the control children are added.

## Changes

- Log what was changed
- Log what was changed
- Log what was changed

## Validation

image
image


petrochuk (approved):

src/Persistence/Yaml/ControlFormatter.cs
```diff
@@ -59,9 +59,9 @@ public static class ControlFormatter
         var isGroupContainer = control.Template.Name == BuiltInTemplates.GroupContainer.Name;
         var originalChildCount = control.Children.Count;
 
-        var children = control.Children
-            .Select(addZIndex)
-            .Concat(childrenToAdd)
+        var children = childrenToAdd
+            .Concat(control.Children
+            .Select(addZIndex))
```
tehcrashxor, line 64: Style Nitpick:  
with the current indentation, the `.Select(addZIndex)` at first glance appears to apply to the result of the concatenation, not just the `control.Children` portion.

It would read clearer with something like one of the below:

```suggestion
            .Concat(control.Children
                .Select(addZIndex))
```
```suggestion
            .Concat(control.Children.Select(addZIndex))
```

tehcrashxor (approved):

tehcrashxor (approved):

PR #706, opened 2024-07-26
Add custom exception for errors from Yaml Validator Library

abaskk-msft:
- Added custom exception to handle expected exceptions from YamlValidator. This will be used in PAC CLI to indicate the error came from the YamlValidator library if the schema couldn't be loaded from the assembly when the IValidatorFactory is asked to create a new validator.

- Also changed the printing format of Validator Results (inconsequential)

## Problem
Custom exception to be used by library callers to handle errors

## Solution

## Changes

- Log what was changed
- Log what was changed
- Log what was changed

## Validation

### DIFF
--- a/src/YamlValidator/YamlValidatorLibraryException.cs
+++ b/src/YamlValidator/YamlValidatorLibraryException.cs
@@ -1,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+public class YamlValidatorLibraryException : Exception
+{
+    public string Reason { get; }
+    public YamlValidatorLibraryException(string reason)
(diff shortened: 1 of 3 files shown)


petrochuk (approved):

src/YamlValidator/YamlValidatorLibraryException.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.YamlValidator;
+public class YamlValidatorLibraryException : Exception
+{
+    public string Reason { get; }
```
mizrael, line 7: use the `Message` property from the base `Exception`

mizrael (changes requested):

mizrael (approved):

PR #714, opened 2024-09-10
added detection for duplicate names in yaml sequence

mizrael:
Added specific error code and detection for duplicate names in yaml sequence.
Example:
```
- MyControl:
    Control: Label
- MyControl:
    Control: Label
```

### DIFF
--- a/src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
+++ b/src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
@@ -115,6 +115,10 @@ public static class PaYamlSerializer
 
             return value;
         }
+        catch (YamlException ex) when (ex.InnerException is ArgumentException && ex.InnerException.Message.Contains("An item with the same key has already been added"))
+        {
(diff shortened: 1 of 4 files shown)


src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
```diff
@@ -115,6 +115,10 @@ private static void WriteTextWriter<TValue>(TextWriter writer, in TValue? value,
 
             return value;
         }
+        catch (YamlException ex) when (ex.InnerException is ArgumentException && ex.InnerException.Message.Contains("An item with the same key has already been added"))
```
joem-msft, line 118: note: I purposely left this special handling out as it's assuming things which could produce incorrect error messages.
But this change is simple enough to be changed later iif needed.

joem-msft, line 118: Instead of adding another catch clause, you can change the existing one to compute the error code based on the where clause.

joem-msft (approved):

tehcrashxor (approved):

PR #716, opened 2024-09-13
Updated DataSources schema to add ConnectorId and update the schema definition.

jamevaalet:
ConnectorId will be an optional property to differentiate between different types of connectors and will have default value as "Dataverse" if not supplied.

## Solution

## Changes

- Added optional ConnectorId property to DataSource Instance OM
- Updated tests to reflect this change
- Updated the yaml schema definition to reflect this change.

## Validation

- local testing


src/schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -435,24 +435,42 @@ definitions:
...
-          TableLogicalName:
-            type: string
-            minLength: 1
+            const: Table
+          Parameters:
+            type: object
+            properties:
+              TableLogicalName:
+                type: string
+            additionalProperties: false
```
joem-msft, line 467: nit: I put the 'additionalProperties' before the proeprties since the properties is often quite long on some types. just makes it easier to find and consistent with the rest of the file

src/schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -435,24 +435,42 @@ definitions:
...
+      - required: [Type]
+        additionalProperties: false
+        properties:
+          Type:
+            const: Actions
+          additionalProperties: false
```
joem-msft, line 473: remove, I think this was a copy-paste error.

src/schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -435,24 +435,42 @@ definitions:
   DataSource-name:
```
joem-msft, line 435: great job editing this file!
after you get done, you need to update the distributed version, which removes comments and such but is intended to be consumed by users.

Do this by running: **src\schemas\publish.cmd**

joem-msft (changes requested):

schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -401,24 +401,41 @@ definitions:
...
     properties:
-      Type: { $ref: "#/definitions/DataSource-Type" }
+      Type:
+        type: string
+        enum:
+          - Table
+          - Actions
+      ConnectorId:
+        type: string
+        nullable: true
```
joem-msft, line 417: sorry I missed this. I don't think 'nullable' is valid in json schema. I've noticed that sometimes the vs code redhat ext suggests patterns in the json schema that's not valid.

For what you want to validate, it's best to just state that it should be a string, and never null.

schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -401,24 +401,41 @@ definitions:
...
-            const: DataverseTable
-          TableLogicalName:
-            type: string
-            minLength: 1
+            const: Table
+          Parameters:
+            type: object
+            additionalProperties: false
+            properties:
+              TableLogicalName:
```
joem-msft, line 432: is this a required property? if so, add: `required: [TableLogicalName]` before line 430.

joem-msft (approved):

schemas/pa-yaml/v3.0/pa.schema.yaml
jamevaalet: It's only required when the Type is Table

joem-msft: correct, which is what this line is doing.

joem-msft: looks like the latest iteration is all correct.

joem-msft (approved):

joem-msft (approved):

PR #717, opened 2024-09-14
ControlInstance type now supports versions

joem-msft:
- Rename `ControlInstance.ControlTypeId` to `ControlType`  
   Now that we'll be supporting control type versions, renaming this property keeps it's semantics from getting confused with the actual `ControlTypeIdentifier`
- Updated schema to allow for version numbers

### DIFF
--- a/src/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/src/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -126,10 +126,11 @@ definitions:
     $comment: Defines reusable schema for validating the pattern allowed for control type identifiers.
     type: string
     # The value of this identifier takes the format:
-    #   ( <invariantFamily> '/' )? <invariantName>
+    #   ( <invariantFamily> '/' )? <invariantName> ( '@' <version> )?
     #   Where the set of allowed characters is limited.
+    #   Where <version> is a 3-part version number (e.g. "1.20.30").
     pattern: |-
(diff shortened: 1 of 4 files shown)


src/schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -126,10 +126,11 @@ definitions:
     $comment: Defines reusable schema for validating the pattern allowed for control type identifiers.
     type: string
     # The value of this identifier takes the format:
-    #   ( <invariantFamily> '/' )? <invariantName>
+    #   ( <invariantFamily> '/' )? <invariantName> ( '@' <version> )?
     #   Where the set of allowed characters is limited.
+    #   Where <version> is a 3-part version number (e.g. "1.20.30").
     pattern: |-
-      ^([A-Z][a-zA-Z0-9]*/)?[A-Z][a-zA-Z0-9]*$
+      ^([A-Z][a-zA-Z0-9]*/)?[A-Z][a-zA-Z0-9]*(@\d+\.\d+\.\d+)?$
```
jamevaalet, line 133: What is the default version when version is not specified?

jamevaalet (approved):

src/schemas/pa-yaml/v3.0/pa.schema.yaml
joem-msft: it's not specified in the schema. This change just indicates the format for the string that is expected.
In the compiler is where a default version will get applied. When no version is set, it'll imply the current version of the control.

tehcrashxor (approved):

PR #720, opened 2024-09-18
added method to detect if a yaml fragment is a sequence

mizrael:
Added a new `CheckIsSequence` method to detect if a yaml fragment is a sequence. It internally creates a new YamlDotNet deserializer that uses the custom type converter `YamlSequenceTesterConverter`.
This converter will simply check if the YAML fragment starts with the `SequenceStart` event and ends with `SequenceEnd`.

The following benchmark shows the results of a few deserialization methods:
- Direct deserialization to `IEnumerable<object>` using a basic YamlDotNet deserializer
- Direct deserialization to `object[]` using a basic YamlDotNet deserializer
- Direct deserialization to `Dictionary<string,object>[]` using a basic YamlDotNet deserializer
- the newly introduced `CheckIsSequence` method
- deserialization using the v2 deserializer
- deserialization using the v3 deserializer
image

results from `CheckIsSequence` are not so different from Direct deserialization to `IEnumerable<object>`. We might decide to remove the custom `YamlSequenceTesterConverter` in the future and rely directly on the base library instead.

### DIFF
--- a/src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
+++ b/src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
@@ -128,5 +128,45 @@ public static class PaYamlSerializer
         // See: fluentvalidation.net
         // See: https://www.youtube.com/watch?v=jblRYDMTtvg
     }
+
+    /// <summary>
+    /// checks the input source is a sequence of YAML items.
+    /// </summary>
+    /// <param name="yaml">The YAML text to check.</param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentNullException"></exception>
+    public static bool CheckIsSequence(string yaml)
+    {
+        _ = yaml ?? throw new ArgumentNullException(nameof(yaml));
+
+        using var reader = new StringReader(yaml);
+        return CheckIsSequence(reader);
+    }
+
+    /// <summary>
+    /// checks the input source is a sequence of YAML items.
+    /// </summary>
+    /// <param name="reader"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentNullException"></exception>
+    public static bool CheckIsSequence(StringReader reader)
+    {
+        ArgumentNullException.ThrowIfNull(reader);
+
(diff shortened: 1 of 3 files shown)


src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
```diff
@@ -128,5 +128,45 @@ private static void WriteTextWriter<TValue>(TextWriter writer, in TValue? value,
...
+        var builder = new DeserializerBuilder()
+            .IgnoreUnmatchedProperties()
+            .WithTypeConverter(new YamlSequenceTesterConverter());
+        var deserializer = builder.Build();
+        try
+        {
+            var results = deserializer.Deserialize<object[]>(reader);
+            return results is not null;
+        }
+        catch (YamlException)
```
joem-msft, line 165: we should not need to use exceptions as control-flow.

Another design option would be to deseralize into a custom type, register a converter for that type, and in the converter populate a property which has the 'result' of the detection.

e..g
```cs
private record SequenceDetectorResult
{
  public required bool IsSequence { get; init; }
}
```

src/Persistence/PaYaml/Serialization/YamlSequenceTesterConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+    public bool Accepts(Type type)
+    {
+        return type == typeof(IEnumerable) || type.IsSubclassOf(typeof(IEnumerable)) ||
+            type == typeof(IEnumerable<object>) || type.IsSubclassOf(typeof(IEnumerable<object>)) ||
+            type == typeof(object[]);
+    }
+
+    public object? ReadYaml(IParser parser, Type type)
+    {
+        if (parser.Current is not SequenceStart)
```
joem-msft, line 25: when using the `SequenceDetectorResult` pattern I mentioned above, this converter could just have this SequenceStart check, completely skip the contents, and then return the result value.

the benefit here is that there wouldn't be an exception thrown for code flow control.

joem-msft (approved):

PR #721, opened 2024-09-18
improved sequence check

mizrael:
updated implementation after feedback received in https://github.com/microsoft/PowerApps-Tooling/pull/720

image


src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
```diff
@@ -153,14 +154,15 @@ public static bool CheckIsSequence(StringReader reader)
...
         {
-            var results = deserializer.Deserialize<object[]>(reader);
-            return results is not null;
+            var parser = new Parser(reader);
+
+            // Need to consume the StreamStart and DocumentStart events to get to the root sequence
+            parser.TryConsume<StreamStart>(out _);
+            parser.TryConsume<DocumentStart>(out _);
+
+            return parser.TryConsume<SequenceStart>(out _);
```
tehcrashxor, line 165: Hmm.   This would certainly be faster for checking if the start of string is a sequence (what I'd expect to see maybe 99.9% of the time), but wouldn't this report a false positive claiming that the following is a sequence?
```yaml
- foo
- bar
asdf: 'jkl;'
```

mizrael: yes. I was discussing this with @joem-msft . I guess the point of the method is to detect a potential sequence, not to validate the correctness of the yaml fragment. That would be left to the deserializer later on.

I'm open to suggestions though. The previous method is slower (even though we're talking microseconds here...), but safer.

joem-msft: the above yaml is invalid, 'asdf' is a mapping key, but the previous lines indicate a sequence.
This scenario is out of scope of the utility. We only want to do the quickest check on the input to determine whether it looks like a sequence.

calling code will still need to deserialize the input to a type.

tehcrashxor: Yeah that string is invalid yaml, but it is a possible fragment to receive by someone copy/pasting but missing the first line, like if the full string was something like 
```yaml
properties:
- foo
- bar
asdf: 'jkl;'
```
or 
```yaml
button1:
  properties:
  - foo
  - bar
  asdf: 'jkl;'
```

but if we're doing other validation for the invalid yaml later down the chain, then the point is moot.  ¯\\\_(ツ)_/¯

tehcrashxor (approved):

src/Persistence.Tests/PaYaml/Serialization/PaYamlSerializerTests.cs
```diff
@@ -214,5 +214,15 @@ public void IsSequenceCheckShouldWorkAsExpected(string path, bool expected)
         isSequence.Should().Be(expected);
     }
 
+    [TestMethod]
+    [DataRow("")]
+    [DataRow("   ")]
+    [DataRow("not yaml")]
```
joem-msft, line 220: this IS yaml, it's a string.

src/Persistence/PaYaml/Serialization/PaYamlSerializer.cs
```diff
@@ -146,21 +150,25 @@ public static bool CheckIsSequence(string yaml)
...
+    /// <exception cref="ArgumentNullException">thrown when <paramref name="reader"/> is null.</exception>
     public static bool CheckIsSequence(StringReader reader)
     {
         ArgumentNullException.ThrowIfNull(reader);
 
-        var builder = new DeserializerBuilder()
-            .IgnoreUnmatchedProperties()
-            .WithTypeConverter(new YamlSequenceTesterConverter());
-        var deserializer = builder.Build();
         try
```
joem-msft, line 163: I don't think we should need to put the try/catch here as the code uses the TryConsume methods only, which shouldn't throw

src/Persistence.Tests/PaYaml/Serialization/PaYamlSerializerTests.cs
```diff
@@ -214,5 +214,15 @@ public void IsSequenceCheckShouldWorkAsExpected(string path, bool expected)
...
+    public void IsSequenceCheckShouldReturnFalseWhenYamlInvalid(string yaml)
```
joem-msft, line 221: also add tests for inline syntaxes:
- `[]`
- `['str`,333]`
- `{ prop1: str2 }`

joem-msft (changes requested):

joem-msft (approved):

PR #723, opened 2024-09-20
Add layout to V3 schema

swapnilbhalgat:
## Problem
For form controls, layout can be edited by makers to change the visual layout of the control. Hence, we need to serialize it with YAML schema

## Solution
Add layout to V3 schema

## Changes
Adds layout to V3 schema

### DIFF
--- a/src/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/src/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -230,6 +230,9 @@ definitions:
         Variant:
           type: string
           minLength: 1
+        Layout:
+          type: string
+          minLength: 1
         Group: true
         Properties: true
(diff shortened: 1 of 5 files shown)


src/schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -230,6 +230,9 @@ definitions:
         Variant:
           type: string
           minLength: 1
+        Layout:
```
joem-msft, line 233: this is correct!
To finalize, you need to manually run: `src\schemas\publish.cmd` and it'll compile and update the published schema. check that change in too.

joem-msft (commented):
this looks all good to go. just run the src\schemas\publish.cmd.

joem-msft (approved):

PR #728, opened 2024-11-27
Use TestMethodAttribute and use MSTest.Sdk

### LINKED ISSUE #726: Switch all `[DataTestMethod]` to `[TestMethod]`
See https://github.com/microsoft/testfx/issues/4166

Youssef1313:
Fixes #726 + uses MSTest.Sdk


Evangelink (commented):
Few notes:

1. You haven't enabled MSTest runner on all projects
2. You will need to update the pipeline setup to use new arguments
3. For now the new runner doesn't support coverlet through `coverlet.collector` so you can remove the deps on all projects. If owners are fine, we can switch to MS code coverage otherwise as part of the pipeline update we will need to move to coverlet dotnet tool

Evangelink (approved):

Youssef1313:
@joem-msft Can I get a review on this PR please?

.github/workflows/CI.yml
```diff
@@ -40,11 +40,8 @@ jobs:
       run: ./build ci
       if: ${{ runner.os != 'Windows' }}
 
-    - name: Test - Persistence
-      run: dotnet bin/Debug/Persistence.Tests/Persistence.Tests.dll
-
-    - name: Test - PAModel
-      run: dotnet bin/Debug/PAModelTests/PAModelTests.dll
```
Youssef1313, line 47: @joem-msft FYI the assembly YamlValidator.Tests.dll wasn't run in CI. It's better to `dotnet test` on the solution rather than listing individual dlls.

.github/workflows/CI.yml
```diff
@@ -40,11 +40,8 @@ jobs:
...
+    - name: Test
+      run: dotnet test src/PASopa.sln --no-build
```
Evangelink, line 44: For maintainers: I would recommend to enable TRX report and Code Coverage.

Evangelink (approved):

mesganmsft (dismissed):

Youssef1313:
@mesganmsft Thanks for reviewing! Is this good to merge now?

.github/workflows/CI.yml
joem-msft: please do not start including the YamlValidator.Tests in the CI builds, as this project has been abandoned and not maintained.

.github/workflows/CI.yml
joem-msft: go ahead and enable if you would like.

Youssef1313: Will follow-up on this as it wasn't already enabled previously.

.github/workflows/CI.yml
Youssef1313: Done.

PR #731, opened 2024-12-18
serialize Locked

swapnilbhalgat:
## Problem
Certain controls show up as locked in the Studio. We need to serilaize Locked with YAML code to implement this feature.

### DIFF
--- a/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -212,6 +212,8 @@ definitions:
         Layout:
           type: string
           minLength: 1
+        Locked:
+          type: boolean
         Group: true
         Properties: true
         Children: { $ref: "#/definitions/Children-Control-instance-sequence" }
--- a/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -212,7 +212,7 @@ definitions:
         Layout:
           type: string
           minLength: 1
-        Locked:
+        IsLocked:
           type: boolean
         Group: true
         Properties: true
--- a/src/schemas-tests/pa-yaml/v3.0/Examples/Src/Screens/FormsScreen2.pa.yaml
+++ b/src/schemas-tests/pa-yaml/v3.0/Examples/Src/Screens/FormsScreen2.pa.yaml
@@ -46,6 +46,7 @@ Screens:
                       Children:
                         - "First Name_DataCard1":
                             Control: TypedDataCard
+                            Locked: true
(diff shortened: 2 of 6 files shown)


schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -212,6 +212,8 @@ definitions:
         Layout:
           type: string
           minLength: 1
+        Locked:
```
mesganmsft, line 215: use "IsLocked"

swapnilbhalgat: Spoken with Joe and Marcel and decided to use Locked since it is shorter and we do not have other boolean properties. I have renamed it to IsLocked for now, we can finalize tomorrow.

src/schemas-tests/pa-yaml/v3.0/Examples/Src/Screens/FormsScreen2.pa.yaml
```diff
@@ -65,6 +66,7 @@ Screens:
                             Children:
                               - DataCardKey7:
                                   Control: Label
+                                  IsLocked: true
```
joem-msft, line 69: only the parent data card should get the IsLocked property set.
the code at design time in the studio applies the final IsLocked value by looking whether it has a parent that is locked.

make sure our test code keeps this semantic.

joem-msft (changes requested):

src/schemas-tests/pa-yaml/v3.0/Examples/Src/Screens/FormsScreen2.pa.yaml
```diff
@@ -46,6 +46,7 @@ Screens:
                       Children:
                         - "First Name_DataCard1":
                             Control: TypedDataCard
+                            IsLocked: true
```
joem-msft, line 49: nit: can we put IsLocked after the Control and Variant properties?

joem-msft: this will keep it consistent with how the object model would serialize it along with keeping type-specific keywords before other optional keywyrs

joem-msft (approved):
nit

src/schemas-tests/pa-yaml/v3.0/Examples/Src/Screens/FormsScreen2.pa.yaml
joem-msft: to be clear, move to after Variant, but before Properties and Children

mesganmsft (approved):

joem-msft (approved):

PR #732, opened 2024-12-19
Downgrade system text json version

swapnilbhalgat:
## Problem
Downgrade system text json version to 8.0.5 since there was a version conflict between the version used by DocServer

### DIFF
--- a/src/PAModel/packages.lock.json
+++ b/src/PAModel/packages.lock.json
@@ -19,9 +19,9 @@
       },
       "System.Text.Encodings.Web": {
         "type": "Direct",
-        "requested": "[9.0.0, )",
-        "resolved": "9.0.0",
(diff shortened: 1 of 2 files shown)


src/PAModel/packages.lock.json
```diff
@@ -30,16 +30,15 @@
       },
       "System.Text.Json": {
         "type": "Direct",
-        "requested": "[9.0.0, )",
-        "resolved": "9.0.0",
-        "contentHash": "js7+qAu/9mQvnhA4EfGMZNEzXtJCDxgkgj8ohuxq/Qxv+R56G+ljefhiJHOxTNiw54q8vmABCWUwkMulNdlZ4A==",
+        "requested": "[8.0.0, )",
+        "resolved": "8.0.0",
```
tehcrashxor, line 34: Hmm.  Despite the csproj explicitly asking for 8.0.5, looks like the lockfile is still grabbing 8.0.0?

tehcrashxor (approved):

PR #737, opened 2025-02-13
Fix OM to allow detection of missing keywords

joem-msft:
In order to allow the PaYamlParser to detect when a keyword is missing, we need to ensure value types are allowed to be set to null. Also, to reduce memory footprint, collections are allowed to be null too.


swapnilbhalgat (approved):

joem-msft:
In order to allow the PaYamlParser to detect when a keyword is missing, we need to ensure value types are allowed to be set to null. Also, to reduce memory footprint, collections are allowed to be null too.

PR #740, opened 2025-03-14
Add `MetadataKey` keyword to control instances

joem-msft:
## Problem

For dataCard and typedDataCard instances, we must enable setting a `MetadataKey` which will be mapped to a `metadataId` in the control template's variant so that changing layouts will work.

## Validation

- Added unit tests coverage.

### DIFF
--- a/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -209,6 +209,9 @@ definitions:
         Variant:
           type: string
           minLength: 1
+        MetadataKey:
+          type: string
+          minLength: 1
         Layout:
           type: string
           minLength: 1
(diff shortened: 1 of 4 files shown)


schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -209,6 +209,9 @@ definitions:
         Variant:
           type: string
           minLength: 1
+        MetadataKey:
```
jamevaalet, line 212: How will the user know what MetadataKey should be used? Are we planning to publish a list?

jamevaalet (approved):

tehcrashxor (approved):

swapnilbhalgat (approved):

schemas/pa-yaml/v3.0/pa.schema.yaml
joem-msft: this'll need to be added to documentation, which we really don't have atm.

PR #743, opened 2025-03-27
Add EditorState to PA YAML schema to support editor UI ordering

jamevaalet:
## Summary
This PR adds support for the `EditorState` top-level property to the PA YAML schema files. EditorState provides editor-specific metadata that doesn't affect app functionality but improves the editing experience by preserving UI element ordering.

## Details
- Added `EditorState` as a top-level property in the schema
- Created `EditorState-instance` definition with two properties:
  - `ScreensOrder`: An ordered array of screen names for editor display
  - `ComponentDefinitionsOrder`: An ordered array of component definition names for editor display
- Schema changes align with existing C# model classes (`PaModule` and `EditorStateInstance`)
- Applied changes to both schema locations:
  - `/src/schemas/pa-yaml/v3.0/pa.schema.yaml`
  - `/schemas/pa-yaml/v3.0/pa.schema.yaml`

## Benefits
- Enables proper schema validation for EditorState in .pa.yaml files
- Provides IntelliSense for EditorState properties in VS Code
- Allows preserving UI tree order across editing sessions
- Maintains a clean separation between functional app logic and editor UI preferences

## Testing
Verified schema validation with existing example files in `/src/schemas-tests/pa-yaml/v3.0/Examples/Src/_EditorState.pa.yaml`.

## Related work
This schema change supports the existing C# model implementation in:
- `src/Persistence/PaYaml/Models/SchemaV3/PaModule.cs`
- `src/Persistence/PaYaml/Models/SchemaV3/EditorStateInstance.cs`

### DIFF
--- a/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -16,6 +16,8 @@ properties:
     $ref: "#/definitions/ComponentDefinitions-name-instance-map"
   DataSources:
     $ref: "#/definitions/DataSources-name-instance-map"
+  EditorState:
+    $ref: "#/definitions/EditorState-instance"
 
 defaultSnippets:
   - label: App
@@ -502,6 +504,22 @@ definitions:
       $ref: "#/definitions/entity-property-name"
     additionalProperties:
       $ref: "#/definitions/pfx-formula"
+
+  EditorState-instance:
+    description: Represents editor state information that's not part of the app logic but helps with editing experience.
+    type: object
+    additionalProperties: false
+    properties:
+      ScreensOrder:
+        description: Ordered list of screen names representing the order in the editor.
+        type: array
+        items:
+          $ref: "#/definitions/Screen-name"
+      ComponentDefinitionsOrder:
+        description: Ordered list of component definition names representing the order in the editor.
+        type: array
+        items:
+          $ref: "#/definitions/ComponentDefinition-name"
 
   entity-name:
     description: The base requirements for a named entity in an app.
--- a/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -17,7 +17,7 @@ properties:
   DataSources:
     $ref: "#/definitions/DataSources-name-instance-map"
   EditorState:
-    $ref: "#/definitions/EditorState-instance"
+    $ref: "#/definitions/EditorState"
 
 defaultSnippets:
   - label: App
@@ -505,7 +505,7 @@ definitions:
     additionalProperties:
       $ref: "#/definitions/pfx-formula"
 
-  EditorState-instance:
+  EditorState:
     description: Represents editor state information that's not part of the app logic but helps with editing experience.
     type: object
     additionalProperties: false
--- a/schemas/pa-yaml/v3.0/pa.schema.yaml
+++ b/schemas/pa-yaml/v3.0/pa.schema.yaml
@@ -506,7 +506,7 @@ definitions:
       $ref: "#/definitions/pfx-formula"
 
   EditorState:
-    description: Represents editor state information that's not part of the app logic but helps with editing experience.
+    description: Represents metadata related to the editor state of the application that enhances the editing experience.
     type: object
     additionalProperties: false
     properties:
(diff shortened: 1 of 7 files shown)


schemas/pa-yaml/v3.0/pa.schema.yaml
```diff
@@ -16,6 +16,8 @@ properties:
     $ref: "#/definitions/ComponentDefinitions-name-instance-map"
   DataSources:
     $ref: "#/definitions/DataSources-name-instance-map"
+  EditorState:
+    $ref: "#/definitions/EditorState-instance"
```
joem-msft, line 20: this isn't an 'instance'. rename to just "EditorState".

src/Persistence/PaYaml/Models/SchemaV3/EditorStateInstance.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.SchemaV3;
+
+public record EditorStateInstance
+{
+    public IList<string>? ScreensOrder { get; init; }
```
joem-msft, line 10: for public records and serialization types, don't use interface types. they also make usage harder as the new `[]` like syntax doesn't work.

use `List<string>` or `string[]` or `ImmutableArray<string>`.
I don't see an existing need in the schema OM to store a list of primitives, so not sure there's precedence.

src/Persistence/PaYaml/Models/SchemaV3/PaModule.cs
```diff
@@ -10,6 +10,8 @@ public record PaModule
 {
     public AppInstance? App { get; init; }
 
+    public EditorStateInstance? EditorState { get; init; }
```
joem-msft, line 13: maybe we should put this after all the known object types? as the last property.

src/schemas-tests/pa-yaml/v3.0/FullSchemaUses/EditorStateSample.pa.yaml
```diff
@@ -0,0 +1,74 @@
...
+    Properties:
+      Height: =100
+      Width: =250
+    Children:
+      - Label1:
+          Control: Label
+          Properties:
+            Text: ="Component 1"
+
+# Editor state defines the visual order in the tree view
```
joem-msft, line 63: I don't expect these comments will survive round-tripping. resist adding documentation to these FullSchemaUses test files.

joem-msft (changes requested):

src/Persistence/PaYaml/Models/SchemaV3/EditorStateInstance.cs
jamevaalet: ImmutableArray<string> sounds more appropriate since the contents should remain immutable after parsing.

joem-msft (approved):

marcelbf (approved):

src/schemas/pa-yaml/v3.0/pa.schema.yaml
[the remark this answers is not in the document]
joem-msft: the schema in the `src/schemas` folder is the source code for the schema.
the one at the root is the built schema. which has dev comments removed, along with some other modifications which are helpful when developing the schema.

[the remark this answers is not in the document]
joem-msft: I wish we had a build process instead that would submit the schema up to some non-git-hub schema store. but that'll have to come when we want to make the tooling story better.

src/Persistence/PaYaml/Models/SchemaV3/EditorStateInstance.cs
```diff
@@ -1,13 +1,13 @@
...
 
-using System.Collections.Generic;
+using System.Collections.Immutable;
 
 namespace Microsoft.PowerPlatform.PowerApps.Persistence.PaYaml.Models.SchemaV3;
 
 public record EditorStateInstance
 {
-    public IList<string>? ScreensOrder { get; init; }
+    public ImmutableArray<string>? ScreensOrder { get; init; }
```
joem-msft, line 10: when using ImmutableArray, don't use `?`. Instead set it to have a default value.
`public ImmutableArray<string> ScreensOrder { get; init; } = [];`

this will ensure no nullreferenceexceptions.

joem-msft (changes requested):

src/Persistence/PaYaml/Models/SchemaV3/EditorStateInstance.cs
jamevaalet: updated to string[] since YamlDotNet can't hande ImmutableArray

joem-msft (approved):

PR #744, opened 2025-04-08
Unit Test for UI Bug YAML De-Serialization Not Working As Expected

madhurima09nath:

## Problem

Writing Unit Test for LoadingPage.OnVisible in shows as escaped string in codeview

## Solution

Added unit test method to make this issue a known bug. Verified the code is functioning as is seen in UI.

## Changes

Added a unit test method to class PFxExpressionYamlConverterTests.cs

## Validation

Unit Test Validation

### DIFF
--- a/src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
+++ b/src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
@@ -137,37 +137,23 @@ public class PFxExpressionYamlConverterTests : SerializationTestBase
     {
         using var _ = new AssertionScope();
 
-        VerifySerialize(
-        """" 
-        Set(gblFoo, true);
-        // The next line is completely blank:
+        VerifySerialize("Set(gblFoo, true);\n // The next line is completely blank:\n\nSet(gblBar, 42)",
+            "|-\n  =Set(gblFoo, true);\n   // The next line is completely blank:\n\n  Set(gblBar, 42)");
 
-        Set(gblBar, 42);
-        """",
-        """"
-        |-
-          =Set(gblFoo, true);
-          // The next line is completely blank:
-
-          Set(gblBar, 42);
-        """");
-
-        // BUG: YamlDotNet doesn't serialize string when spaces are added in blank lines and shows escaped string in codeview
+        // BUG:31691580 YamlDotNet doesn't serialize string when spaces are added in blank lines and shows escaped string in codeview
         // LoadingPage.OnVisible with spaces on blank line in the middle, serialization adds \r\n for new lines on windows-os and \n for non-windows os
 
-        VerifySerialize(
-        """"
-        Set(gblFoo, true);
-        // The next line is completely blank:
-          
-        Set(gblBar, 42);
-        """",
-        """
-        "=Set(gblFoo, true);\r\n// The next line is completely blank:\r\n  \r\nSet(gblBar, 42);"
-        """
-        );
+        VerifySerializeWithEscapedStringInCodeView("Set(gblFoo, true);\r\n// The next line has one or more spaces:\r\n    \r\nSet(gblBar, 42);",
+            """=Set(gblFoo, true);\r\n// The next line has one or more spaces:\r\n    \r\nSet(gblBar, 42);""");
+    }
 
-
+    private void VerifySerializeWithEscapedStringInCodeView(string? pfxScript, string expectedExpressionYaml)
+    {
+        var expression = pfxScript is null ? null : new PFxExpressionYaml(pfxScript);
+        var testObject = new NamedPFxExpressionYaml(expression);
+        var expectedYaml = expectedExpressionYaml is null ? "Expression:" : $"Expression: \"{expectedExpressionYaml}\"\n";
+        var actualYaml = SerializeViaYamlDotNet(testObject);
+        actualYaml.Should().Be(expectedYaml);
     }
 
     private void VerifySerialize(string? pfxScript, string expectedExpressionYaml)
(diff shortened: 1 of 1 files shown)


src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
```diff
@@ -132,6 +132,44 @@ public void WriteYamlWithTrailingWhitespace()
...
+        """",
+        """"
+        |-
+          =Set(gblFoo, true);
+          // The next line is completely blank:
+
+          Set(gblBar, 42);
+        """");
+
+        // BUG: YamlDotNet doesn't serialize string when spaces are added in blank lines and shows escaped string in codeview
```
joem-msft, line 155: add the bug number to the comment. e.g. `// BUG 12345: description here`

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
```diff
@@ -132,6 +132,44 @@ public void WriteYamlWithTrailingWhitespace()
...
+        // LoadingPage.OnVisible with spaces on blank line in the middle, serialization adds \r\n for new lines on windows-os and \n for non-windows os
```
joem-msft, line 156: Is this line is applicable to this test or the actual bug.

note: one caveot of the raw string `"""` is that the line returns might be inconsistent on different OSes, dependent on the build machine.
I would consider adding a `.Replace("\r\n", "\n")` or vice versa to the strings to ensure the inputs have the NewLine chars expected for the test.

joem-msft: otherwise, you could always go back to using non-raw strings as the input, and be very explicit about the input.
I know this goes against my claim to try to use the raw string syntax. So use your own judgement on whether this is applicable.

joem-msft (changes requested):

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
madhurima09nath: resolved.

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
```diff
@@ -132,6 +132,30 @@ public void WriteYamlWithTrailingWhitespace()
...
+    {
+        using var _ = new AssertionScope();
+
+        VerifySerialize("Set(gblFoo, true);\n // The next line is completely blank:\n\nSet(gblBar, 42)",
+            "|-\n  =Set(gblFoo, true);\n   // The next line is completely blank:\n\n  Set(gblBar, 42)");
+
+        // BUG:31691580 YamlDotNet doesn't serialize string when spaces are added in blank lines and shows escaped string in codeview
+        // LoadingPage.OnVisible with spaces on blank line in the middle, serialization adds \r\n for new lines on windows-os and \n for non-windows os
+
+        VerifySerializeWithEscapedStringInCodeView("Set(gblFoo, true);\r\n// The next line has one or more spaces:\r\n    \r\nSet(gblBar, 42);",
```
joem-msft, line 146: to simplify the difference here with the case on line 140, ensure this input uses "\n" and not "\r\n".

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
```diff
@@ -132,6 +132,30 @@ public void WriteYamlWithTrailingWhitespace()
...
+            """=Set(gblFoo, true);\r\n// The next line has one or more spaces:\r\n    \r\nSet(gblBar, 42);""");
```
joem-msft, line 147: is the raw string literal necessary for this?
Instead, use regular string, and ensure you escape the double quotes if they're expected.
e.g. `"\"=Set(..., 42);\""`

joem-msft: oic, you're calling a new function. the new function is only different by the expectedExpressionYaml being surrounded by double quotes. In actuality, this will lead to potentially invalid escaping of the expected string if someone tries to use it in the future.

I think this make the code more complex w/o saving much code. It only saves 4 chars in the calling code.

Here's an example where (I think) this new method would fail:
```cs
VerifySerializeWithEscapedStringInCodeView("Set(gblFoo,\n   \n   \"Hello world!\");",
"Set(gblFoo,\n   \n   \"Hello world!\");");
```

consider making call to existing `VerifySerialize` like so:
```cs
VerifySerialize("Set(gblFoo, true);\n// The next line has one or more spaces:\n    \nSet(gblBar, 42);",
"\"=Set(gblFoo, true);\n// The next line has one or more spaces:\n    \nSet(gblBar, 42);\"");
```
or with my example:
```cs
VerifySerialize("Set(gblFoo,\n   \n   \"Hello world!\");",
"\"Set(gblFoo,\n   \n   \\\"Hello world!\\\");\""); // note the extra encoding of the backslash
```
(I may be wrong with my example here as to the correct expected string, but my comment here is to use the existing method.

This allows the caller to own the expected yaml encoding.

joem-msft (changes requested):

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
madhurima09nath: discussed and resolved.

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
madhurima09nath: Added Hello World example. Discussed and resolved.

joem-msft (approved):

PR #751, opened 2025-08-12
Force literal block for Unicode characters to preserve them without escaping

shshetty-microsoft:
Force literal block for Unicode characters to preserve them without escaping

## Problem

The Unicode characters were being escaped.

## Solution

Force literal block for Unicode characters to preserve them without escaping

### DIFF
--- a/src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
+++ b/src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
@@ -151,6 +151,48 @@ public class PFxExpressionYamlConverterTests : SerializationTestBase
             "\"=Set(gblFoo, true);\\n// The next line has one or more spaces:\\n    \\nSet(gblBar, 42);\"");
     }
 
+    [TestMethod]
+    public void WriteYamlWithUnicodeCharacters()
+    {
+        using var _ = new AssertionScope();
+
+        // Test Unicode characters in PowerFx expressions are preserved
+        VerifySerialize("Set(用户名, \"John\")", "|-\n  =Set(用户名, \"John\")");
+        VerifySerialize("Text(\"こんにちは世界\")", "|-\n  =Text(\"こんにちは世界\")");
+        VerifySerialize("Format(价格, \"$#,##0.00\")", "|-\n  =Format(价格, \"$#,##0.00\")");
+
+        // Test mixed Unicode and ASCII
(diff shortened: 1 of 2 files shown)


src/Persistence/PaYaml/Serialization/PFxExpressionYamlConverter.cs
```diff
@@ -58,6 +58,8 @@ public void WriteYaml(IEmitter emitter, object? value, Type type)
         // Force multi-line scripts to be literal blocks
         forceLiteralBlock |= expression.InvariantScript.Contains('\n')
             || expression.InvariantScript.Contains('\r');
+        // Force literal block for Unicode characters to preserve them without escaping
+        forceLiteralBlock |= expression.InvariantScript.Any(c => c > 127 && !char.IsControl(c));
```
tehcrashxor, line 62: Is the issue occurring for every character above 2F / 127?
For instance, is that occurring for the €, ¥, £ (Euro, Yen, British Pound) found in the 80 - FF (128-255) range?

src/Persistence.Tests/PaYaml/Serialization/PFxExpressionYamlConverterTests.cs
```diff
@@ -151,6 +151,48 @@ public void WriteYamlWithEscapedStringInCodeView()
             "\"=Set(gblFoo, true);\\n// The next line has one or more spaces:\\n    \\nSet(gblBar, 42);\"");
     }
 
+    [TestMethod]
+    public void WriteYamlWithUnicodeCharacters()
+    {
+        using var _ = new AssertionScope();
+
+        // Test Unicode characters in PowerFx expressions are preserved
+        VerifySerialize("Set(用户名, \"John\")", "|-\n  =Set(用户名, \"John\")");
```
joem-msft, line 160: to remove the need to escape quotes in these test inputs, use the new C# syntax: `"""Set(用户名, "John")"""`
the """ in a row can be done inline or a multiline string. But for these tests I think a single line is going to be fine.
the benefit is that the inputs need less C# escape chars.
do this for all of your intput strings in the new test methods.

joem-msft: not this will only work if your input string doesn't also start with a " char.

src/Persistence/PaYaml/Serialization/PFxExpressionYamlConverter.cs
joem-msft: this is a good point. maybe we should have a more targeted conditional?

PR #808, opened 2026-06-02
deprecate YamlValidator

joem-msft:
- Removing YamlValidator out of build process as it's deprecated.
- Removed old PaYamlV2 schema files now deprecated.


docs/YAMLFileFormat.md
```diff
@@ -1,13 +1,13 @@
-# Legacy Power Apps YAML Source File Format
+# Experimental Power Apps YAML Source File Format
 
 > [!WARNING]
-> **This format is generated by PASopa and will likely change**
+> **This Experimental format is generated by PASopa and will likely change**
```
tehcrashxor: Without extra context, I would assume something labeled "experimental" would be the *newer* option.  
Might want to make it clearer that this is the deprecated route that won't recieve support.

tehcrashxor (approved):

tehcrashxor (approved):

PR #809, opened 2026-06-02
PAModel - modernize to langVer 13

joem-msft:
Now that we're expecting to address some Experimental layout bugs, we're going to improve the assembly to utilize modern coding practices.
- Updates the LangVersion to 13, matching Persistence lib
- Add reference to `PolySharp` to fill framework gaps
- Add reference to Persistence lib so we can start sharing code, e.g. TfmAdapters, TfmExtensions
- Addressed some low-hanging fruit language constructs available in latest C# language


src/PAModel/PAConvert/Yaml/YamlLexer.cs
```diff
@@ -419,7 +418,7 @@ private YamlToken Unsupported(LineParser line, string message)
 
     private YamlToken UnsupportedSingleQuote(LineParser line, bool isComponent)
     {
-        var lineSplit = line._line.ToString().ToLower().Split(new string[] { " as " }, StringSplitOptions.None);
+        var lineSplit = line.Line.ToString().ToLower().Split([" as "], StringSplitOptions.None);
```
tehcrashxor: nit - skip the array initialization with the overload that takes `string`

src/PAModel/PAConvert/Yaml/YamlWriter.cs
```diff
@@ -32,7 +32,7 @@ public void WriteStartObject(string propertyName)
     {
         WriteIndent();
 
-        var needsEscape = propertyName.IndexOfAny(new char[] { '\"', '\'' }) != -1;
+        var needsEscape = propertyName.IndexOfAny(['\"', '\'']) != -1;
```
tehcrashxor: nit - consider storing as a `private static readonly char[]` to prevent re-allocating the array every time.  (Same below.)

Might also be able to use `SearchValues<char>` depending on if the earlier frameworks handel that with the polyfil

tehcrashxor (approved):

src/PAModel/PAConvert/Yaml/YamlLexer.cs
joem-msft: that overload doesn't exist for net48. but I'll add a wrapper for it.

src/PAModel/PAConvert/Yaml/YamlWriter.cs
joem-msft: net48 is making things annoying. I've improved it a little though.

tehcrashxor (approved):

